Issue #2, opened 2018-05-04
TASK: Create symbol address list


RadWolfie:
We need to create a generated list for each symbol, not each revision, somehow. Either macro or existing structure method.

Then we can use it in Cxbx-Reloaded project to make references for patch and trampoline to avoid workaround we have at the moment.

LukeUsher:
The list generation and HLE Cache file generation can remain in Cxbx-Reloaded, as they are function specific.

I was thinking that the Symbol Scanning engine would remain in this project, outside of Cxbx-Reloaded, so that it could be called in this manner

Cxbx-Reloaded code:
```C++
std::map<std::string, xbaddr> g_SymbolAddresses;
void SymbolCallback(char* symbolName, xbaddr symbolAddress) {
  g_SymbolAddresses[symbolName] = symbolAddress;
}

XbSymbolScan(SymbolCallback);  // Call into this library
CxbxApplyPatches(); // Iterate through g_SymbolAddresses & resolve patches 
```

And everything else would remain in this project.
The ScanForXboxSymbols function would be part of this library and simply take a function pointer as input, and call it for every symbol found.

That way, the actual method of using the symbols is generic, and the application that uses this library can make all decisions on how to store or use the data returned.

RadWolfie:
You finally found the best solution! I'm glad you thought about this before I lose my hair. 😉

LukeUsher:
I'll be doing some work on Cxbx-Reloaded this weekend to decouple scanning and patching, so HLEIntercept/HLEDatabase will be renamed to SymbolDatabase and SymbolScan (so they can easily be removed later, once this project is mature) and there will be new function files for HLEPatch. 

That way, the symbol scanning and patching will become entirely separate. Sorry it took so long to get here!

PatrickvL:
I'm afraid one thing is overlooked : the fact that symbol scanning is using XREF's to 'harden' searches.

An XREF is a pre-defined enumeration, that appears in a few forms : 
* as an ID for an OOVPA (registered in the XREF table once an OOVPA is found)
* as derived XREF's (currently hard-coded from a select few D3D OOVPA) 
* as cross-checked XREF's (XREF's which have to be found already, and which have to match the potential hit)

Once symbol scanning is completed, it's important to not only return the located OOVPA-based symbol locations, but also include all XREF symbol address. How's that going to play together with the idea of a symbol-callback? Call it afterwards for all XREF's? But by what name? Please discuss...

[...]

PR #12, opened 2018-05-04
Update #2

RadWolfie:
* Removed global variable access.
* Removed internal functions access.
* Renamed GetSymbolDataBaseHash to XbSymbolLibraryVersion.
* Add XbSymbol scan functions
* Fixed few warnings

Currently missing XREF storing I think. It will temporarily break until we have internal functions working as is. It is a step forward to getting what we need for the API prepare.

### DIFF
--- a/XbSymbolDatabase.c
+++ b/XbSymbolDatabase.c
@@ -118,7 +118,6 @@ const unsigned int SymbolDBListCount = OOVPA_TABLE_COUNT(SymbolDBList);
 unsigned int XRefDataBase[XREF_COUNT] = { 0 }; // Reset and populated by EmuHLEIntercept
 
 bool bXRefFirstPass = false; // For search speed optimization, set in EmuHLEIntercept, read in EmuLocateFunction
-uint32_t UnResolvedXRefs = 0; // Tracks XRef location, used (read/write) in EmuHLEIntercept and EmuLocateFunction
 
 
 // ******************************************************************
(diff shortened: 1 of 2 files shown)


RadWolfie:
For time being when if @LukeUsher agree with the API arguments he needs for his plugin. We can merge this and continue the fixes internally.

XbSymbolDatabase.c
```diff
@@ -247,6 +311,76 @@ extern uint32_t EmuLocateFunction(OOVPA *Oovpa, uint32_t lower, uint32_t upper)
...
+                XRefDataBase[Oovpa->XRefSaveIndex] = address;
+                register_func((void*)0, OovpaTable->szFuncName, address, OovpaTable->Version);
+            }
+        } else {
+            register_func((void*)0, OovpaTable->szFuncName, address, OovpaTable->Version);
+        }
+    }
+}
+
+bool XbSymbolScan(void* xbeData, char* section_name, uint32_t lower_bound, uint32_t upper_bound, xb_symbol_register_t register_func) {
```
RadWolfie: I made a discovery, void* xbeData isn't needed as lower_bound and upper_bound arguments are only two things we need. Since these are the memory addresses.

PatrickvL: There's a difference between the buffer pointer which has to be scanned, and the virtual address where this buffer should reside.

These two can be identical, but in the case of an IDA Pro plugin, the buffer data won't be loaded at the actual image base address, hence the need of the two.

Finally, we either need the start of the buffer, or an ending address (either the end of the buffer, or the end of virtual memory).

If I may express my thoughts, I would pass on the arguments as : buffer + size + image_base

PatrickvL: Uncertainties like this are exactly the reason behind my prior proposal to pass in only one argument - a struct that we can modify more easily than the function arguments themselves.

A few questions/remarks on the arguments you listed :
* uint32_t xbe_address, // is this the address of the buffer to be scanned, or it's virtual base address?
* uint32_t xbe_size, // if this size applies to the argument above, then it's fine either way
* char* section_name, // why is this needed?
* uint32_t section_address, // why pass in a section address at all?
* uint32_t section_size, // why would you need the size of the section? 
* _symbol_register_t register_func // this callback might be simpler with only argument too

RadWolfie: * uint32_t xbe_address, // Address in memory, it can be anywhere in memory.
* uint32_t xbe_size, // Correct, this is part of "XBE_MAX_VA" usage.
* char* section_name, // We can only scan specific section for accuracy reasons. Please, don't argue about this.
* uint32_t section_address, // This is the "lower_bound".
* uint32_t section_size, // This is part of "upper_bound". Instead of re-adjust upper_bound the program itself, we can input what's already existed in xbe's header.
* _symbol_register_t register_func // Already has a simple function arguments contain everything we need now and the future. This is in lockdown.

PatrickvL: Thanks for adding more detail. Do note that I'm not so much arguing about this, as instead getting a clear picture of what the API needs. Right now there's at least one thing not described well enough (it's still ambiguous) : what's the relation between Xbe address and section address?

RadWolfie: Section address is where XbSymbolDB library start scanning with OOVPA database along with . xbe_address and xbe_size is only use to enforce XREF range check similar method with XBE_MAX_VA, you're using.

PR #15, opened 2018-05-06
Fix warnings part 2

### LINKED ISSUE #7: TASK: Fix warnings
Currently there are repeat warnings for OOVPAs. For now it does compile and tested with Cxbx-Reloaded as functional.

However, we still need to fix it anyway.

RadWolfie:
Final warning messages are now gone.

Resolve #7


RadWolfie:
Since this is only for warning issues fix, merging into master branch.

And yes I am aware this may cause breakage for coding. It will be fix in next update.

PR #30, opened 2018-06-08
Ignore warning for nonstandard extension from Visual Studio

RadWolfie:
It works as intended plus flexible array member is supported since C99 standard.

The warning from Visual Studio is `C4200: nonstandard extension used: zero-sized array in struct/union`.


RadWolfie:
Low risk pull request, merging now.

PR #32, opened 2018-06-12
More updates

RadWolfie:
This pull request contain...
* Treat warnings as errors. (since there are absolutely no warning anymore)
* Check xbe header for library section contain DSOUND string before enforce scan for DSOUND symbols from known 4039 missing library version data.
* Replace goto to "do while" statement.
* Fix coding to our standard formatting by hand.
* Remove some extra checking in if statements.
* Ignore certain spots in D3D checking if below 3911 when supposed to.
  * We don't support any unofficial titles under revision 3911 builds in HLE.


RadWolfie:
Merging since it is working fine on my end.

PR #33, opened 2018-06-26
Fix false positive in Indigo Prophecy

LukeUsher:
This symbol (3911 D3DDevice_MultiplyTransform)  was being detected incorrectly in the .text section of Indigo Prophecy (5849)

I added a couple of extra bytes to the 3911 signature to avoid this false positive. It still matches the 3911 version as expected (Tested with THPS 2X)


LukeUsher:
Force merging this to fast-track: It solves a regression that needs to be fixed before v0.1 can go live

RadWolfie:
Verify both 3911 and 5849 titles with same symbol OOVPA signature match at known address: ✔️
Verify any other OOVPAs after 3911 signature: ❌ (There is 4039 OOVPA signature.)
Verify titles between 3911 and 5849 to remain valid: ❔

If 3911 signature is matching 5849 and there is newer OOVPA signature, 4039 OOVPA. It needs to be tested for 3911 signature to match for 4039 title.

These are the reason why at least one other person skilled with X86 and OOVPAs database should be checking this before approval to merge. 😛

LukeUsher:
@RadWolfie I tested with 5 titles each using a different XDK, and MultiplyTransform was correctly found in them all. I can't recall exactly what games I used, but I definitely used 3911, something 4XXX, 5849 and some other inbetweens.

If you find another false detection, document it and we'll fix it.

I fast-tracked this because it fixed a *major* regression in Indigo Prophecy, a title that otherwise works very well. It *needed* to be in v0.1 before it goes live

Issue #42, opened 2018-09-05
Wrong recognition


gandalfthewhite19890404:
Situation:
Some titles (GTA San Andreas for example) may refuse to boot and switches to dashboard with that screen:
![image](https://user-images.githubusercontent.com/24764335/45077165-22ac6680-b0f5-11e8-947e-9a23d45d826c.png)

But on recent builds it throws error:
![image](https://user-images.githubusercontent.com/24764335/45077194-36f06380-b0f5-11e8-8783-b3350c3cde29.png)

Crash caused by these two extra lines in HLE/Symbol cache (for example - San Andreas):
D3DDevice_SelectVertexShader_0=35c664
D3DDevice_SetShaderConstantMode_0=35c714

HLE Cache revision is same - 716367461.

jarupxx:
I was confirmed. Fixed by PR #43.
These functions are match standard argument signatures and LTCG specific signatures, so It was necessary to add retn opcode.

gandalfthewhite19890404:
Fixed, closing issue.

PR #43, opened 2018-09-18
OOVPA : Various fixes D3D8 OOVPAs

jarupxx:
This pull request includes
- Fix issues #42. False positive in D3DDevice_SelectVertexShader_0 and D3DDevice_SetShaderConstantMode_0.
Fix false positive by adding retn opcode. However, possibly false negative due the LTCG signature changes length.
- Separate D3D8 4039 database.
- Add D3D8LTCG OOVPAs.


PatrickvL:
Looks good to me!

RadWolfie:
Hmm... something is off with 4039 titles. `D3DDevice_SetRenderTarget` is missing from this pull request's generated symbol cache file. I believe this pull request is valid since `D3DDevice_SetRenderTarget` was moved and remain unchanged. I'm requiring at least couple days to investigate source of cause before re-validate the pull request.

jarupxx:
I have corrected the D3DDevice_SetRenderTarget regression you pointed out. Thank you.

RadWolfie:
I'm a little concerned with starting offset only has one instead of usual of 2 or more for better lockdown. For now, I am approving this pull request since you have tested both non-LTCG and LTCG titles.

PR #44, opened 2018-10-02
Register D3DRS_CullMode

LukeUsher:
I came up with a solution to auto-detect and fix an incorrect detection of EmuD3DDeferredRenderState within Cxbx-Reloaded, however, it requires that the offset of the derived D3DRS_CULLMODE symbol is passed to the emulator.

This PR makes that happen.

The Cxbx-R side will be pushed shortly: Just note that Cxbx-R uses this information to automatically correct a broken EmuD3DDeferredRenderState, as well as show a notification to submit an issue to fix it upstream in XbSymbolDatabase: This is beneficial for everyone.


PatrickvL:
LGTM

RadWolfie:
Resolved by commit cc2b1b5

PR #45, opened 2018-10-22
Cleanup for non-Windows platfoms and non-MSVC compilers

JayFoxRox:
- Fix case-sensitive include-path: Unlike Windows, some platforms are case-sensitive. So the include path must be written properly
- Fix use of the `inline` keyword: The inline keyword has a different meaning in different versions of C. I turned the functions which are only in one compilation unit into `static inline`. The function which isn't accessible from elsewhere isn't `inline` anymore.

The code is untested, because there's many other issues in the codebase, and I don't have 32 bit compatibility to test XbSymbolDatabase at all (See issue 25 about x64 support).

### DIFF
--- a/XbSymbolDatabase.c
+++ b/XbSymbolDatabase.c
@@ -203,7 +203,7 @@ const char* XbSymbolLibraryToString(uint32_t library_flag)
     }
 }
 
-inline uint32_t XbSymbolLibrayToFlag(const char* library_name)
+uint32_t XbSymbolLibrayToFlag(const char* library_name)
 {
     if (strncmp(library_name, Lib_D3D8, 8) == 0) {
         return XbSymbolLib_D3D8;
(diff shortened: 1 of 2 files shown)


RadWolfie:
Sorry about the case-sensitive include paths. Either I might had overlooked it when I was correcting some or my own manual type-in.

So, using inline XbSymbolLibrayToFlag didn't work? Only reason I had inline there is for optimization.

JayFoxRox:
The `inline` keyword changed meaning in different versions of the C language and with extensions.

My understanding is that `inline` just means "don't export this symbol". [So if it's `inline`, then you suddenly get "Undefined reference" errors in some dialects (I think I tried GNU in my case, with a default CMake project)](https://stackoverflow.com/questions/19068705/undefined-reference-when-calling-inline-function).

Using `static inline` does get rid of this linking-error, and is probably what the author intended.
The functions are meant to be static anyway, so it shouldn't change anything.

In the case of the function which is declared in the header, and visible outside the module, this wouldn't work (the function can't be static). I'm not sure how to use `inline` for those functions or if the language even supports it. You'd need LTO for that anyway. Hence I've removed the `inline` to get rid of the linking-error.

Also, it's worth nothing that regardless of these hint keywords, the function is likely to be inlined by the compiler on higher optimizations (which is also typically a condition for LTO to begin with).

RadWolfie:
> My understanding is that inline often just means "don't export this symbol"

True, I trust you with this. Merging as is. 😉

Issue #25, opened 2018-05-11
Add support for x64 program


RadWolfie:
I noticed some warning outputs for x64 build by accident from AppVeyor's build.
```
XbSymbolDatabase.c(284): warning C4312: 'type cast': conversion from 'uint32_t' to 'uint32_t *' of greater size [C:\projects\cxbx-reloaded\import\XbSymbolDatabase\xbSymbolDatabase.vcxproj]
...
XbSymbolDatabase.c(655): warning C4312: 'type cast': conversion from 'unsigned int' to 'unsigned int *' of greater size [C:\projects\cxbx-reloaded\import\XbSymbolDatabase\xbSymbolDatabase.vcxproj]
...
XbSymbolDatabase.c(1104): warning C4267: 'function': conversion from 'size_t' to 'const unsigned int', possible loss of data [C:\projects\cxbx-reloaded\import\XbSymbolDatabase\xbSymbolDatabase.vcxproj]
```

If we need add support for such as IDA Pro's plugin feature. It's something we'll have to do to fix.

PatrickvL:
Xbox code will at most use 32 bit memory address, so any address or offset can be stored with a 32 bit number. But 64 bit plugins must be prepared to access actual buffer addresses above 32 bit. So that should use either void pointer, or uintptr_t

JayFoxRox:
In my opinion, most of the code should still only use 32-bit addresses. The re-mapping from Xbox to host space should occur in separate address translation / virtual memory read functions.

RadWolfie:
Much as I would like to stay specific 32-bit addresses. Except there would not be a good support for 64-bit's 4GB higher range address. The scanning process has to be able support it. What we could do is hard-coded found xbox's 32-bit address to callback's register function.

JayFoxRox:
I merely meant that the address calculation for Xbox addresses should remain 32-bit. I'm fine with translating addresses for buffer lookups to some other type: `void* translate_xbox_address(uint32_t xbox_virtual_address[, size_t byte_count])` (as final step).
There could also be memory accessor functions like `uint32_t read_xbox_u32(uint32_t xbox_virtual_address)`.

This translation function could be supplied as a callback (to make it the users responsibility), or be part of XbSymbolDatabase (assuming that the memory layout is non-paged and XBE virtual memory image is entirely loaded into memory with ImageSize bytes, at whatever base the user wants).
I prefer it be part of XbSymbolDatabase for simplicity.

[...]

PR #46, opened 2018-10-22
Use fixed-size types in Xbe.h

JayFoxRox:
A first step towards #25.

This will require "stdint.h" (C99+).
I did not want to mess with `char` types. We should have a static assert for those.

The code is untested, because there's many other issues in the codebase, and I don't have 32 bit compatibility to test XbSymbolDatabase at all. I did not even confirm correctness of the Xbe.h struct layouts on their own.

### DIFF
--- a/Xbe.h
+++ b/Xbe.h
@@ -25,107 +25,109 @@
 // ******************************************************************
 #pragma once
 
+#include <stdint.h>
+
 #pragma pack(1)
 
 // Xbe section header
 typedef struct _xbe_s_flags {
-    unsigned int bWritable : 1;                 // writable flag
-    unsigned int bPreload : 1;                  // preload flag
-    unsigned int bExecutable : 1;               // executable flag
-    unsigned int bInsertedFile : 1;             // inserted file flag
-    unsigned int bHeadPageRO : 1;               // head page read only flag
-    unsigned int bTailPageRO : 1;               // tail page read only flag
-    unsigned int Unused_a1 : 1;                 // unused (or unknown)
-    unsigned int Unused_a2 : 1;                 // unused (or unknown)
-    unsigned int Unused_b1 : 8;                 // unused (or unknown)
-    unsigned int Unused_b2 : 8;                 // unused (or unknown)
-    unsigned int Unused_b3 : 8;                 // unused (or unknown)
+    uint32_t bWritable : 1;                     // writable flag
+    uint32_t bPreload : 1;                      // preload flag
+    uint32_t bExecutable : 1;                   // executable flag
+    uint32_t bInsertedFile : 1;                 // inserted file flag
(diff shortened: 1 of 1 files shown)


RadWolfie:
Part of the reason I do not use typedef to allow compile with fixed-size native method. As I investigate more, apparently 16-bit's int is 2 byte size for some reason. I'll accept this. However... the comments are not align after the changes you made. Can you fix this or should I do it for you?

JayFoxRox:
> Part of the reason I do not use typedef to allow compile with fixed-size native method.

Sorry, I can't follow. What do you mean?
I don't use any `typedef` here myself, but it wouldn't cause any issues either.

Most compilers and architectures don't care about data type and have instructions or patterns to deal with common sizes (8,16,32 or 64 bit. Often also 24 or 56 bit for float compatibility).

> As I investigate more, apparently 16-bit's int is 2 byte size for some reason. 

Isn't it expected that 16-bit integers are 2 bytes? If you meant "16-bit platform / arch", then this is an unlikely case. The C types usually have a minimum size requirement (which should be 16-bit for `int`).
So it's very possible some platform will eventually have larger integers (typically 32-bit at the moment, but possibly larger).

*I was actually under the assumption that clang was already using 64-bit `int` when writing the code. This did not hold true, but I still think it's a valid change.*

> However... the comments are not align after the changes you made. Can you fix this or should I do it for you?

I'll do it later or sometime in the next days. Good catch. I didn't pay attention to that.

RadWolfie:
> Sorry, I can't follow. What do you mean?
> I don't use any `typedef` here myself, but it wouldn't cause any issues either.

stdint.h are using typedef types by the way. 😛

> Isn't it expected that 16-bit integers are 2 bytes? If you meant "16-bit platform / arch", then this is an unlikely case. The C types usually have a minimum size requirement (which should be 16-bit for `int`).
> So it's very possible some platform will eventually have larger integers (typically 32-bit at the moment, but possibly larger).
> 
> _I was actually under the assumption that clang was already using 64-bit `int` when writing the code. This did not hold true, but I still think it's a valid change._

Both of your statements are correct. Your code changes brought it up when I was not aware of integers are not fixed-size back then. However, 32-bit program using `int` are 4 bytes afaik with Microsoft's Visual Studio. Except 64-bit program are indeed using 8 bytes (64-bit integer of course). My original belief is using `long` is 4 bytes for 32-bit and 8 bytes for 64-bit which had been carry over for years until maybe a year ago. Such a shame compiler can be very picky without stdint.h's typedefs.

> I'll do it later or sometime in the next days. Good catch. I didn't pay attention to that.

Thanks! 😃

JayFoxRox:
I've updated the code with aligned comments.

Please verify that the code still works for you, as I don't have any 32-bit environment to test it.

PR #47, opened 2018-11-23
Batch code optimize

RadWolfie:
Contains:
* Migrate much as possible between D3D8 and D3D8LTCG codes into one.
  * To reduce duplicate codes as possible.
* Small fix up for 64-bit support.
   * Doesn't cover all of it and still doesn't support outside of xbox address range.
* 1 minor bug with duplicate for statements.
  * programmer bug (oops)

### DIFF
--- a/XbSymbolDatabase.c
+++ b/XbSymbolDatabase.c
@@ -125,10 +125,10 @@ const unsigned int SymbolDBListCount = OOVPA_TABLE_COUNT(SymbolDBList);
 // ******************************************************************
 // * XRefDataBase
 // ******************************************************************
-unsigned int XRefDataBase[XREF_COUNT] = { 0 }; // Reset and populated by EmuHLEIntercept
+uint32_t XRefDataBase[XREF_COUNT] = { 0 }; // Reset and populated by EmuHLEIntercept
 
 bool bXRefFirstPass; // For search speed optimization, set in XbSymbolScan, read in XbSymbolLocateFunction
-unsigned int UnResolvedXRefs = XREF_COUNT;
+uint32_t UnResolvedXRefs = XREF_COUNT;
 
 
 // ******************************************************************
@@ -152,6 +152,7 @@ void XbSymbolSetOutputMessage(xb_output_message_t message_func)
 {
     output_func = message_func;
 }
+
 void XbSymbolOutputMessage(xb_output_message mFlag, const char* message) {
     if (output_func != 0) {
         output_func(mFlag, message);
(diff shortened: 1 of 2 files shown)


XbSymbolDatabase.c
```diff
@@ -971,15 +971,12 @@ bool XbSymbolScan(void* xbeData, xb_symbol_register_t register_func)
         bool bDSoundLibHeader = false;
 
         // Verify if title do contain DirectSound library section.
-        for (unsigned int v = 0; v < pXbeHeader->dwSections; v++) {
+        for (uint32_t v = 0; v < pXbeHeader->dwSections; v++) {
```
JayFoxRox: should stick to `unsigned int` here. It's guaranteed to be large enough but gives the compiler more options to work with.

XbSymbolDatabase.c
```diff
@@ -305,8 +306,8 @@ bool CompareOOVPAToAddress(OOVPA *Oovpa, uint32_t cur)
 
 // locate the given function, searching within lower and upper bounds
 uint32_t XbSymbolLocateFunction(OOVPA *Oovpa,
-                                uint32_t lower,
-                                uint32_t upper)
+                                size_t lower,
+                                size_t upper)
```
JayFoxRox: Not sure if `size_t` is such a great choice:

If this is C99+ anyway, then we have stdint, which provided `uintptr_t` for an integer large enough to hold a pointer.

However, for user-facing APIs, this should be using `void*` (also available pre-C99) for host addresses; and `uint32_t` (or a `typedef` of such) for Xbox addresses.

RadWolfie: Hmm... I don't see uintptr_t in stdint header file from Window's end. However, I do find it in vadefs.h header file.

uintptr_t seems to be what I was originally looking for both 32 and 64 bit memory range compatible. I'll replace it soon.

XbSymbolDatabase.c
RadWolfie: Okay, I'm usually habit with size bound to avoid screwing up the uneven size problem in the past.

RadWolfie:
No more additions to this pull request. I have another branch ready for another pull request.

RadWolfie:
Typedef of xbaddr or hwaddr should be done in separate pull request, I understand what you were saying.

I will review about void* later. Currently, the code does not affect unaligned size between 32 and 64 bit program.

RadWolfie:
> @JayFoxRox However, for user-facing APIs, this should be using void* (also available pre-C99) for host addresses; and uint32_t (or a typedef of such) for Xbox addresses.

For scanning process, void* type can't be used due to increment/decrement usage. Honestly, it's better off with uintptr_t type to make things easier. Passing down the pointer of host memory from front-end can remain as-is.

> @PatrickvL Anything related to host pointers can be declared using "{type}*" or "void*" just fine.

I'm sure it apply the same to above response.

Beside, I don't believe there is any side effect just for using uintptr_t type internally.

PR #48, opened 2018-11-24
Support 64-bit program and outside hardcode xbox base memory range

### LINKED ISSUE #25: Add support for x64 program

RadWolfie:
Finally, 64-bit program and no restriction on memory range are supported.

using XbSymbolScan function is deprecated. Recommendation are:
* XbSymbolScanFileImage (full file image scan purpose)
* XbSymbolScanRunTime (scan runtime xbe by an emulator or the like)

resolve #25

P.S. Any extended concern or found something off (anything really). Feel free to comment below.

### DIFF
--- a/XbSymbolDatabase.c
+++ b/XbSymbolDatabase.c
@@ -125,10 +125,10 @@ const unsigned int SymbolDBListCount = OOVPA_TABLE_COUNT(SymbolDBList);
 // ******************************************************************
 // * XRefDataBase
 // ******************************************************************
-uint32_t XRefDataBase[XREF_COUNT] = { 0 }; // Reset and populated by EmuHLEIntercept
+xbaddr XRefDataBase[XREF_COUNT] = { 0 }; // Reset and populated by EmuHLEIntercept
 
 bool bXRefFirstPass; // For search speed optimization, set in XbSymbolScan, read in XbSymbolLocateFunction
-uint32_t UnResolvedXRefs = XREF_COUNT;
+unsigned int UnResolvedXRefs = XREF_COUNT;
 
 
 // ******************************************************************
@@ -271,7 +271,7 @@ bool CompareOOVPAToAddress(OOVPA *Oovpa, uintptr_t cur, uintptr_t xb_start_virt_
 
         // get currently registered (un)known address
         GetXRefEntry(Oovpa, v, &XRef, &Offset);
-        uint32_t XRefAddr = XRefDataBase[XRef];
+        xbaddr XRefAddr = XRefDataBase[XRef];
         // Undetermined XRef cannot be checked yet
         // (XbSymbolLocateFunction already checked this, but this check
         // is cheap enough to keep, and keep this function generic).
@@ -282,9 +282,9 @@ bool CompareOOVPAToAddress(OOVPA *Oovpa, uintptr_t cur, uintptr_t xb_start_virt_
         if (XRefAddr == XREF_ADDR_DERIVE)
             continue;
 
-        uint32_t ActualAddr = *(uint32_t*)(cur + Offset);
+        xbaddr ActualAddr = *(xbaddr*)(cur + Offset);
         // check if PC-relative or direct reference matches XRef
-        if ((ActualAddr + (uint32_t)(cur - xb_start_virt_addr) + Offset + 4 != XRefAddr) && (ActualAddr != XRefAddr))
+        if ((ActualAddr + (xbaddr)(cur - xb_start_virt_addr) + Offset + 4 != XRefAddr) && (ActualAddr != XRefAddr))
             return false;
     }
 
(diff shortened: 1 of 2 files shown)


XbSymbolDatabase.h
```diff
@@ -415,17 +419,26 @@ void XbSymbolSetOutputMessage(xb_output_message_t message_func);
...
 /// <param name="library_flag">Name of the library in flag.</param>
 /// <param name="symbol_str">Name of the library in symbol string.</param>
-/// <param name="address">TBD.</param>
+/// <param name="address">Return xbox's virtual address.</param>
 /// <param name="revision">Found with specific revision.</param>
-typedef void (*xb_symbol_register_t)(const char* library_str, uint32_t library_flag, const char* symbol_str, uint32_t address, uint32_t revision);
+typedef void (*xb_symbol_register_t)(const char* library_str, uint32_t library_flag, const char* symbol_str, xbaddr address, uint32_t revision);
+
+/// <summary>
+/// To scan symbols in runtime xbe usage.
```
JayFoxRox: What is "runtime"? Does this mean "Virtual" ?

XbSymbolDatabase.h
```diff
@@ -415,17 +419,26 @@ void XbSymbolSetOutputMessage(xb_output_message_t message_func);
...
 
 /// <summary>
-/// To scan symbols for Xbe's file.
+/// To scan symbols in Xbe's file image.
 /// </summary>
 /// <param name="xbeData">Starting point of xbe memory address.</param>
 /// <param name="register_func">Callback register function to be call for any detected symbols.</param>
 /// <returns>Only return false if something is not valid.</returns>
-bool XbSymbolScan(const void* xbeData, xb_symbol_register_t register_func);
+bool XbSymbolScanFileImage(const void* xbeData, xb_symbol_register_t register_func);
```
JayFoxRox: What does this do? An Image can be addressed in various ways (Raw / Virtual).

When I read the PR description, I thought this would just load an XBE from file path. But this seems to do something else? So this looks poorly named.

XbSymbolDatabase.h
RadWolfie: Runtime definition: https://techterms.com/definition/runtime

As in emulator usage (even if it's not align to xbox address, it will work). Or a mockup program to load xbe into memory, load specific sections, then perform a scan.

XbSymbolDatabase.h
RadWolfie: `Pointer to Raw image` is what this function do. Plus this function provide a complete scan than XbSymbolScanRunTime function which only check which section is loaded then scan those only.

RadWolfie:
Also, I don't believe file image is the correct term for xbe file. Since it is a executable file, not a iso type nor is a picture format.

My terms of image would be pictures or creating an image of disk/disc (aka iso, img, vhd, etc).

XbSymbolDatabase.c
```diff
@@ -1037,7 +1067,27 @@ bool XbSymbolScan(const void* xbeData, xb_symbol_register_t register_func)
 
                 if (bXRefFirstPass) {
                     if ((LibraryFlag & (XbSymbolLib_D3D8 | XbSymbolLib_D3D8LTCG)) > 0) {
-                        XbSymbolDX8SectionScan(LibraryFlag, pXbeHeader, BuildVersion, LibraryStr, register_func);
+                        for (unsigned int s = 0; s < pXbeHeader->dwSections; s++) {
+                            SectionName = (const char*)(xb_start_addr + pSectionHeaders[s].SectionNameAddr);
+                            if (strncmp(SectionName, Sec_D3D, 8) == 0) {
+
+                                if (isRunTime) {
+                                    // if an emulator did not load a section, then skip the section scan.
```
JayFoxRox: "emulator"? Why is that relevant?

RadWolfie: It's relevant, except it is an xbe executable's responsibility to load a section. 😛

PatrickvL:
https://docs.microsoft.com/en-us/windows/desktop/debug/pe-format says: "This specification describes the structure of executable (image) files and object files under the Windows family of operating systems. "

RadWolfie:
Thanks for the link *and* the quote. Although, I don't agree with Microsoft's term for executable file as image file.

Here's a question we should be asking ourselves, does Unix OSes and Apple call their own executable file as image file? Plus image file in Xbox emulator environment could meant either xiso or xbe file. Putting FileImage in a function can confuse other developers (including myself).

The only thing I agree on is PC Mag's technical term for image file.

RadWolfie:
Merging as-is since it works in XbSymbolCacheGenTest project.

PR #50, opened 2019-01-03
Internal Debug Support

RadWolfie:
More or less a copy and paste from Cxbx-Reloaded's checking for any duplicated OOVPAs and "expanded?" errors. Which also include modification from C++ codes to C. And include error counts for `XbSymbolDataBaseTestOOVPAs` function.

Currently, there are total of 36 errors for D3DDevice_xxxxx symbols.

We also need to include support for verifying different symbol matching same address as well. TBA

NOTE: There may be a chance of add parameter support later on down the path for different test and so on. In theory, it may will be a separate function.

### DIFF
--- a/XbSymbolDatabase.c
+++ b/XbSymbolDatabase.c
@@ -492,7 +492,7 @@ bool XbSymbolScanSection(uint32_t xbe_base_address,
 {
 
     // Invalid argument
-    if (section_name == NULL || xbe_size == 0 || section_size == 0 || register_func == 0) {
+    if (section_name == std_nullptr || xbe_size == 0 || section_size == 0 || register_func == std_nullptr) {
         return 0;
     }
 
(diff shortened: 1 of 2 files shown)


RadWolfie:
PS: We can remove the original debug code from Cxbx-Reloaded. Which I will do that.

PatrickvL:
Looks good to me.

Please tell when you're ready so it can be merged

RadWolfie:
To be honest, we can merge it since the OOVPA progress I'm doing are still work in progress. Yet has benefit for XbSymbolCacheGenTest program to fix OOVPAs sooner.

I'll eventually make a new PR later on for ability to check symbols against duplicated address(es) only from xbe scan.

At least there's a notice for those going to use XbSymbolDataBaseTestOOVPAs function.

PR #51, opened 2019-01-21
Add _CRT_SECURE_NO_WARNINGS to project

RadWolfie:
Since library by itself will not compile unless delcared in other project.

Quick hotfix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,8 @@ cmake_minimum_required (VERSION 3.8)
 
 project (XbSymbolDatabase)
 
+add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
+
 file(GLOB SOURCES "XbSymbolDatabase.c" "OOVPADatabase/*.inl")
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -5,6 +5,8 @@ cmake_minimum_required (VERSION 3.8)
 
 project (XbSymbolDatabase)
 
+add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
```
JayFoxRox: This feels like it should only apply for MSVC

RadWolfie: XbSymbolCacheGenTest has this defined keyword. It's the only way to allow compile with MSVC. Somehow MSVC didn't check XbSymbolDatabase inspection for any defined keyword.

Source: https://github.com/Cxbx-Reloaded/XbSymbolCacheGenTest/blob/master/CMakeLists.txt

RadWolfie: Also, I'm only using MSVC since I am mainly Windows user btw. I cannot vouch for other compiler may have same issue or not.

PR #53, opened 2019-01-28
Add Symbols for Chihiro JVS

LukeUsher:
This successfully detects all JVS Functions in the following Titles:

House of The Dead 3 (4831) 
Crazy Taxi  3 (4831)
Virtua Cop 3 (5028)
Ollie King (5455)
Outrun 2 (5558)
Outrun 2 SP (5659)
Ghost Squad (5788)
Sega Golf 2006 (5849)
SegaBoot v2.00.0

The one exception is JvsRTC_Write, which is not present in any game, but exists within SegaBoot. Signatures for this will be added at a later date, for now, there's not enough test cases.


LukeUsher:
Note that as these are new symbols that previously didn't exist, there is still a possibility these OOVPA will need further tweaking, they have only been tested against a limited subset of titles, so don't take these offsets as fact just yet.

PR #54, opened 2019-01-28
Fix GNU compiler support

JayFoxRox:
This should affect GCC, and probably also Clang (not sure if the preprocessor checks are good).

Commits are self-explanatory:

- `_BitScanReverse` does not exist in the GCC world. The zero-case is never encountered, but I added it for completeness; it returns 32 (which means the first set bit is not within the lower 32 bits).
- `strncat` with `n == sizeof(src)` gives a warning (as zero included - it was still initialized and appended though). There's no good reason to use `strncat` as `strcat` behaves the same. The `strncat` using `OOVPA_TABLE_COUNT` was even worse, as there's no reason this should be done; its only job was a cryptic form of `strlen("string") + 1`.

**Please confirm I did not break Windows support.**
Ideally someone would also test on macOS or using another clang implementation.

*(I did not add myself to the copyright list as this doesn't really have anything that qualifies for that. Please tell me if I should still do it)*

### DIFF
--- a/XbSymbolDatabase.c
+++ b/XbSymbolDatabase.c
@@ -1268,7 +1268,7 @@ void HLEError(SymbolDatabaseVerifyContext *context, uint16_t buildVersion, char 
     }
 
     if (context->against != NULL && context->against_data != NULL) {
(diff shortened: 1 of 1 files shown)


RadWolfie:
Confirmed is able to compile on Windows platform.

PR #55, opened 2019-01-28
Fix detection of JvsScFirmwareDownload and JvsFirmwareDownload

LukeUsher:
Added an XRef for the strings they access. 

This was necessary as these functions differ only in status messages, the actual logic is identical byte-by-byte, so this is the only way to tell the difference


RadWolfie:
Please restore the tab spaces.

OOVPADatabase/Xapi.OOVPA.inl
```diff
@@ -127,57 +127,57 @@
...
-    REGISTER_OOVPAS(XMountMUA, 3911, 4242),
-    REGISTER_OOVPAS(XMountMURootA, 3911, 4242),
-    REGISTER_OOVPAS(XMountUtilityDrive, 3911, 4432),
-    REGISTER_OOVPAS(XRegisterThreadNotifyRoutine, 3911),
-    REGISTER_OOVPAS(XSetProcessQuantumLength, 4134),
-    REGISTER_OOVPAS(XUnmountAlternateTitleA, 3911),
-    REGISTER_OOVPAS(XapiFiberStartup, 3911),
-    REGISTER_OOVPAS(timeKillEvent, 3911),
-    REGISTER_OOVPAS(timeSetEvent, 3911),
+	REGISTER_OOVPAS(CreateMutex, 3911), // Too High Level (from 3911's comment)
```
RadWolfie: Please restore tab spaces

LukeUsher: Ugh, visual studio sucks sometimes, I didn't change anything regarding spacing, it auto-formatted like this when I typed :\

PatrickvL: We could avoid this by adding an editor configuration file to to root of all of our repositories

JayFoxRox: @PatrickvL that quickly escalates to a collection of IDE configs and is generally not a good idea. Stricter review on this is usually enough - it doesn't take much time to check and fix.

There's also even a git filter in https://github.com/Cxbx-Reloaded/XbSymbolDatabase/blob/master/.gitconfig; however, it wasn't used here?

This repository already has an issue with being overly protective: https://github.com/Cxbx-Reloaded/XbSymbolDatabase/blob/master/.gitignore is massive bloat!

Normally such things shouldn't be handled in a projects .gitignore (which is a project-specific file).

For users who use such tools, git also allows a systemwide .gitignore which should be used to hide local files. Unfortunately many users (and maintainers) are unaware and this lead to projects having overly complicated project-files.
What if a new editor appears? Do we just keep adding to this list?

This should be similar for editor configurations: the project shouldn't have a catch-all.

It should just set and enforce the rules: tell users what they should do (and possibly provide simple, universally acceptable tools). For code-style, a single clang-format or checkpatch.pl can be provided, which have become defacto standard for code-formatting (independent of editors). But I don't think it's necessary here - review should be enough.

RadWolfie: @PatrickvL, I had talked to Luke about it and informed him there will be .clang-format to correct the problem.

@JayFoxRox It will be great to have a "whitelist" of what files *should* be included in the git instead of adding various IDEs to it.  Hmm... I found a possibility to do so for have whitelist in .gitignore file. Source: https://stackoverflow.com/questions/8024924/gitignore-ignore-all-files-then-recursively-allow-foo

PR #59, opened 2019-02-15
Organize file path

RadWolfie:
At least the file path is more cleaner to view inside directory than using MSVC's filter view method.

EDIT: This PR will not affect any new/delete/updated OOVPAs. New OOVPA files will need to redirect into right directory if occur in the past.

Plus changes to CMakeLists.txt is optional, yet allow to see complete unfiltered list in IDE. It is possible to remove the optional changes then have developer to open the file via right-click the included file for review, update, etc.


PatrickvL:
The separation in the CMakefile could have been done without any file moves by using this : "OOVPADatabase/D3D8*.inl" (and repeat it for the other libs). But now that you've moved them, it's okay to keep it organized like this.

RadWolfie:
Found the alternate solution which does include ".inl" files in subdirectory.

JayFoxRox:
Skimming over the changes, I'm against this:


- Moving files loses their history/tracking on GitHub (minor issue, at least blame still works). Compare [this](https://github.com/RadWolfie/XbSymbolDatabase/commits/organize-file-path/OOVPADatabase/D3D8/D3D8.r3911.inl) to [this](https://github.com/Cxbx-Reloaded/XbSymbolDatabase/commits/master/OOVPADatabase/D3D8.1.0.3911.inl).
- The old naming scheme was better (1.0.XXXX over rXXXX) because it's what MS actually intended.
- Creates redundancy in the filenames along the lines of `/SameThing/SameThing.r3911.inl`.

So, while having separate folders would have been cleaner, or can be cleaner, I disagree with how this PR restructures it.

> At least the file path is more cleaner to view inside directory than using MSVC's filter view method.

I think this is a bad motivation; What's bad about using the IDE to arrange files?

This can be done through CMake for most IDEs, or MSVC project files, without physically moving files.

While I'd be pro-move, I don't think this has enough benefits to warrant changing the structure.

RadWolfie:
> * Moving files loses their history/tracking on GitHub (minor issue, at least blame still works). Compare [this](https://github.com/RadWolfie/XbSymbolDatabase/commits/organize-file-path/OOVPADatabase/D3D8/D3D8.r3911.inl) to [this](https://github.com/Cxbx-Reloaded/XbSymbolDatabase/commits/master/OOVPADatabase/D3D8.1.0.3911.inl).

So far, only blame has helped preserve the changes, correct. As for put all files in same directory, this was not our original plan. Our intention was to put them in their own directory, which we didn't have the time to do that back then.

> * The old naming scheme was better (1.0.XXXX over rXXXX) because it's what MS actually intended.

We had talked about this before, Patrick and I agree we could remove the `1.0.` *since* there was never a 1.1, 2.0, etc thing. They're just revision builds in original xbox term which is also designed for original xbox title.
> * Creates redundancy in the filenames along the lines of `/SameThing/SameThing.r3911.inl`.

It's useful for tabs viewing before put the OOVPA in the wrong file. Plus we had JVS symbols which is added to Xapi since JVS is part of Xapi for general Chihiro arcade.

JayFoxRox:
> We had talked about this before, Patrick and I agree we could remove the 1.0. since there was never a 1.1, 2.0, etc thing. They're just revision builds in original xbox term which is also designed for original xbox title.

I believe it's not revisions, it's build numbers. So if anything it should be `bXXXX`. The `r` is very confusing in my opinion.

However, *if* you are going to change it, why bother with putting the letter at all? MS also just used the number for short versions. Example: `D3D8.5233.inl` (instead of `D3D8.r5233.inl` or `D3D8.b5233.inl` or ...).

I'd also suggest to avoid double `.` in the filenames. `D3D8_5233.inl` would be cleaner in my opinion.

> It's useful for tabs viewing before put the OOVPA in the wrong file.

Again: bad motiviation.

That just means it's an unclean hack to improve readability in misconfigured IDEs, where the current solution already works fine.

*XQEMU also does this for legacy reasons in components like `nv2a` an `dsp`, but I eventually want to get rid of it, too.*

>  Plus we had JVS symbols which is added to Xapi since JVS is part of Xapi for general Chihiro arcade.

I don't think it's part of XAPI, but a stand-alone general purpose lib by SEGA (not part of XDK).

The comment in the code isn't very precise: `JVS Libraries are linked in the same sections as XAPILIB, so we define them here` *which* section is that? @LukeUsher probably knows.

Regardless, I don't think it's wise to hide the Chihiro functions in XAPI, and the current way of handling it looks cleaner (although the SEGA inherited JVS name is already a tad bit confusing).

RadWolfie:
Using `b` is better, including using underline too. Soon, I'll take a look at `source_group` command for cmake to see if I can do the task. If I can do the task, then JVS will have its own folder.

> I don't think it's part of XAPI, but a stand-alone general purpose lib by SEGA (not part of XDK).

It's part of Chihiro's XDK edition I believe. However, I don't think we ever had a former Chihiro developer to verify it?

Also, some libraries are using the same section name over time. A perfect example would be `XONLINE`, `XONLINES`, `XNET`, `XNETS`.

LukeUsher:
I’m just doing to chime in about the Chihiro JVS function placement. 

Am through the JVS library functions were probably not a part of xapi, their functions generally reside in either the .text or .xpp sections alongside the XInput functions that are part of xapi. 

I would have preferred to add another library definition in the header, but the JVS functions bring no additional library version headers or any additional sections in the executable. Since XbSymbolDatabase uses the library version headers to identify separate libraries, I was unable to make it an entirely separate table, which is why I ended up merging it with xapi.

RadWolfie:
I had put JVS into its own folder. As for add a JVS library header file, it can be done in separate pull request.

Thanks to @JayFoxRox for mention `source_group` command. Commit ca9ed9a introduce ability to create virtual folder for each source files either by open the CMakeLists.txt file or generated the target compiler's files.


Also, ready for final review & approval.

PR #60, opened 2019-02-16
Some Debug Updates

RadWolfie:
Added:
  * XbSymbolBypassBuildVersionLimit to API usage. (See API for details.)
    * Is a requirement in order to perform a depth testing internally.
  * XbSymbolSetOutputVerbosity to API usage. (See API for details.)
    * By include this function, the process will run faster internal by skip certain verbose message with string formats provided.
  * XbSymbolOutputMessageFormat to reduce unnecessary buffer size holder in any functions.
  * Output debug message of what library and section are being scanned.
  * **NEW** XbSymbolContinuousSigScan to API usage. (See API for details.)
    * Is a requirement in order to perform a depth testing internally.
  * **NEW** XbSymbolFirstDetectAddressOnly to API usage. (See API for details.)
    * Is a requirement in order to perform a depth testing internally.

### DIFF
--- a/XbSymbolDatabase.c
+++ b/XbSymbolDatabase.c
@@ -1170,6 +1170,8 @@ bool XbSymbolScan(const void* xb_header_addr,
                                 if (SymbolDBList[d2].LibSec.section[d3] != NULL && strncmp(SectionName, SymbolDBList[d2].LibSec.section[d3], 8) == 0) {
                                     pSectionScan = pSectionHeaders + s;
 
+                                    XbSymbolOutputMessageFormat(XB_OUTPUT_MESSAGE_DEBUG, "Scanning %.8s library in %.8s section", LibraryStr, SectionName);
+
                                     XbSymbolScanOOVPA(SymbolDBList[d2].OovpaTable, SymbolDBList[d2].OovpaTableCount, LibraryStr, SymbolDBList[d2].LibSec.library,
                                                         pSectionScan, BuildVersion, register_func, xb_start_virt_addr);
                                     break;
--- a/XbSymbolDatabase.c
+++ b/XbSymbolDatabase.c
@@ -387,6 +387,7 @@ void* XbSymbolLocateFunction(OOVPA *Oovpa,
         return 0;
 
     uint32_t derive_indices = 0;
+
     // Check all XRefs are known (if not, don't do a useless scan) :
     for (unsigned int v = 0; v < Oovpa->XRefCount; v++) {
         uint32_t XRef;
@@ -419,7 +420,7 @@ void* XbSymbolLocateFunction(OOVPA *Oovpa,
     }
 
     // search all of the image memory
-    for (memptr_t cur = lower; cur < upper; cur++)
+    for (memptr_t cur = lower; cur < upper; cur++) {
         if (CompareOOVPAToAddress(Oovpa, cur, xb_start_virtual_addr)) {
 
             while (derive_indices > 0) {
@@ -455,6 +456,7 @@ void* XbSymbolLocateFunction(OOVPA *Oovpa,
 
             return cur - xb_start_virtual_addr;
         }
+    }
 
     // found nothing
     return 0;
@@ -528,8 +530,9 @@ void XbSymbolScanOOVPA(OOVPATable *OovpaTable,
 
         // Search for each function's location using the OOVPA
         xbaddr pFunc = (xbaddr)(uintptr_t)XbSymbolLocateFunction(pLoop->Oovpa, lower, upper, (uintptr_t)xb_start_virt_addr);
-        if (pFunc == 0)
+        if (pFunc == 0) {
             continue;
+        }
 
         if (pFunc == pLastKnownFunc && pLastKnownSymbol == pLoop - 1) {
             //if (g_SymbolAddresses[pLastKnownSymbol->szFuncName] == 0) {
@@ -544,6 +547,7 @@ void XbSymbolScanOOVPA(OOVPATable *OovpaTable,
         pLastKnownFunc = pFunc;
         pLastKnownSymbol = pLoop;
     }
+
     if (pLastKnownSymbol != NULL) {
         XbSymbolRegisterSymbol(pLastKnownSymbol, LibraryName, LibraryFlag, pLastKnownFunc, register_func);
     }
@@ -848,7 +852,7 @@ void XbSymbolDX8SectionScan(uint32_t LibraryFlag,
         if (BuildVersion < 3911) {
             // Not supported, currently ignored.
         }
-        if (BuildVersion < 4034) {
+        else if (BuildVersion < 4034) {
             pFunc = XbSymbolLocateFunctionCast(&D3DDevice_SetRenderState_CullMode_3911, lower, upper, xb_start_virt_addr);
         }
         else {
--- a/XbSymbolDatabase.c
+++ b/XbSymbolDatabase.c
@@ -469,6 +469,7 @@ void XbSymbolScanOOVPA(OOVPATable *OovpaTable,
     OOVPATable *pLastKnownSymbol = NULL;
     uint32_t pLastKnownFunc = 0;
     const char *SymbolName = NULL;
+    char output[2048];
     for (; pLoop < pLoopEnd; pLoop++) {
 
         if (SymbolName == NULL) {
@@ -494,10 +495,14 @@ void XbSymbolScanOOVPA(OOVPATable *OovpaTable,
 
         if (pFunc == pLastKnownFunc && pLastKnownSymbol == pLoop - 1) {
             //if (g_SymbolAddresses[pLastKnownSymbol->szFuncName] == 0) {
-                char output[2048];
                 sprintf(output, "Duplicate OOVPA signature found for %s, %hd vs %hd!", pLastKnownSymbol->szFuncName, pLastKnownSymbol->Version, pLoop->Version);
                 XbSymbolOutputMessage(XB_OUTPUT_MESSAGE_WARN, output);
             //}
+        }
+
+        if (buildVersion < pLoop->Version) {
+            sprintf(output, "OOVPA signature is too high for [%hd] %s!", pLoop->Version, pLoop->szFuncName);
+            XbSymbolOutputMessage(XB_OUTPUT_MESSAGE_WARN, output);
         }
 
         pLastKnownFunc = pFunc;
--- a/XbSymbolDatabase.c
+++ b/XbSymbolDatabase.c
@@ -187,9 +187,30 @@ void XbSymbolBypassBuildVersionLimit(bool bypass_limit)
     bStrictBuildVersionLimit = !bypass_limit;
 }
 
-void XbSymbolOutputMessage(xb_output_message mFlag, const char* message) {
-    if (output_func != 0) {
-        output_func(mFlag, message);
+// Intended for send the message as-is without formating.
+// (To avoid corrupted string when perform a bad coding.
+//  Plus certain message need special character without format.)
+void XbSymbolOutputMessage(xb_output_message mLevel, const char* message)
+{
+    // Check if output function is set.
+    if (output_func != NULL) {
+        output_func(mLevel, message);
+    }
+}
+
+// Intended to format the message with given extra parameters.
+void XbSymbolOutputMessageFormat(xb_output_message mLevel, const char* format, ...)
+{
+    char bufferTemp[2048];
+
+    // Check if output function is set.
+    if (output_func != NULL) {
+        va_list args;
+        va_start(args, format);
+        (void)vsprintf(bufferTemp, format, args);
+        va_end(args);
+
+        output_func(mLevel, bufferTemp);
     }
 }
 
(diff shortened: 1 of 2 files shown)


XbSymbolDatabase.h
```diff
@@ -482,6 +482,12 @@ uint32_t XbSymbolLibrayToFlag(const char* library_name);
 /// <returns>Return total count of errors.</returns>
 unsigned int XbSymbolDataBaseTestOOVPAs();
 
+/// <summary>
+/// Set to true will perform full range of OOVPAs registered. Or stop at title's revision detected.
```
JayFoxRox: What does "stop at" refer to?
Do you mean "Only scan title revision"?

Is this setting per-library, or what's a "title's revision" (= a generic one across all libs?)

XbSymbolDatabase.h
```diff
@@ -482,6 +482,12 @@ uint32_t XbSymbolLibrayToFlag(const char* library_name);
...
+/// </summary>
+/// <param name="bypass_limit">Input boolean to either bypass or enable the revision limit.</param>
+void XbSymbolBypassRevLimit(bool bypass_limit);
```
JayFoxRox: I have a hard time trying to think of when this should ever be necessary. Can you give a practical example of what you'd debug with this?

XbSymbolDatabase.h
JayFoxRox: If this is primarily a debug feature, this should be mentioned here.

XbSymbolDatabase.c
```diff
@@ -1283,9 +1289,7 @@ void HLEError(SymbolDatabaseVerifyContext *context, uint16_t buildVersion, char
     (void)strcat(buffer, " : ");
     (void)strncat(buffer, bufferTemp, ret_str_count);
 
-    if (output_func != NULL) {
-        output_func(XB_OUTPUT_MESSAGE_ERROR, buffer);
-    }
+    XbSymbolOutputMessage(XB_OUTPUT_MESSAGE_ERROR, buffer);
```
JayFoxRox: Why was this changed?

XbSymbolDatabase.c
```diff
@@ -478,21 +484,21 @@ void XbSymbolScanOOVPA(OOVPATable *OovpaTable,
...
 
-#if 0 // TODO: Find a way to support this.
         if (pFunc == pLastKnownFunc && pLastKnownSymbol == pLoop - 1) {
-            if (g_SymbolAddresses[pLastKnownSymbol->szFuncName] == 0) {
-                printf("HLE: Duplicate OOVPA signature found for %s, %d vs %d!\n", pLastKnownSymbol->szFuncName, pLastKnownSymbol->Version, pLoop->Version);
-            }
+            //if (g_SymbolAddresses[pLastKnownSymbol->szFuncName] == 0) {
+                char output[2048];
+                sprintf(output, "Duplicate OOVPA signature found for %s, %hd vs %hd!", pLastKnownSymbol->szFuncName, pLastKnownSymbol->Version, pLoop->Version);
+                XbSymbolOutputMessage(XB_OUTPUT_MESSAGE_WARN, output);
```
JayFoxRox: Why doesn't this use `output_func` ?

XbSymbolDatabase.c
```diff
@@ -494,12 +495,16 @@ void XbSymbolScanOOVPA(OOVPATable *OovpaTable,
...
         if (pFunc == pLastKnownFunc && pLastKnownSymbol == pLoop - 1) {
             //if (g_SymbolAddresses[pLastKnownSymbol->szFuncName] == 0) {
-                char output[2048];
                 sprintf(output, "Duplicate OOVPA signature found for %s, %hd vs %hd!", pLastKnownSymbol->szFuncName, pLastKnownSymbol->Version, pLoop->Version);
                 XbSymbolOutputMessage(XB_OUTPUT_MESSAGE_WARN, output);
             //}
         }
 
+        if (buildVersion < pLoop->Version) {
+            sprintf(output, "OOVPA signature is too high for [%hd] %s!", pLoop->Version, pLoop->szFuncName);
```
JayFoxRox: nit: If we are using stdint.h (I'm not sure if we do), we should also use inttypes.h: `PRId16` although this should probably be `PRIu16` or `%hu`

XbSymbolDatabase.c
JayFoxRox: Nevermind, `XbSymbolOutputMessage` calls `output_func`. I thought it was the other way around.

It's a bit odd, considering `XbSymbolOutputMessage` looks like an external API, given it's name.

XbSymbolDatabase.c
JayFoxRox: Resolved, [see my other comment](https://github.com/Cxbx-Reloaded/XbSymbolDatabase/pull/60#discussion_r257467231)

XbSymbolDatabase.c
```diff
@@ -187,12 +187,27 @@ void XbSymbolBypassRevLimit(bool bypass_limit)
...
 
-void XbSymbolOutputMessage(xb_output_message mFlag, const char* message) {
+void XbSymbolOutputMessage(xb_output_message mFlag, const char* message)
+{
     if (output_func != 0) {
         output_func(mFlag, message);
     }
 }
 
+void XbSymbolOutputMessageFormat(xb_output_message mFlag, const char* format, ...)
```
JayFoxRox: I don't think this needs to be a separate function, we can just extend `XbSymbolOutputMessage`. The old style of calling it is still valid, and it's not exactly performance critical either.

XbSymbolDatabase.c
```diff
@@ -182,14 +187,24 @@ void XbSymbolSetOutputMessage(xb_output_message_t message_func)
     output_func = message_func;
 }
 
+bool XbSymbolSetOutputVerbose(xb_output_message verbose_level)
```
JayFoxRox: nit: `XbSymbolSetOutputVerbose` &rarr; `XbSymbolSetOutputVerbosity` ?

XbSymbolDatabase.c
```diff
@@ -182,14 +187,24 @@ void XbSymbolSetOutputMessage(xb_output_message_t message_func)
...
+
 void XbSymbolBypassRevLimit(bool bypass_limit)
 {
     bStrictRevisionLimit = !bypass_limit;
 }
 
 void XbSymbolOutputMessage(xb_output_message mFlag, const char* message)
 {
-    if (output_func != 0) {
+    if (output_func != 0 && mFlag >= output_verbose_level) {
```
JayFoxRox: Consider making these early outs for readability (in their respective commits):

```c
    if (output_func == NULL) {
        return;
    }

    if (mFlag < output_verbose_level) {
        return;
    }

    // Do stuff
```

XbSymbolDatabase.c
```diff
@@ -182,14 +187,24 @@ void XbSymbolSetOutputMessage(xb_output_message_t message_func)
...
+    // Return false only when requested level is invalid.
+    return false;
+}
+
 void XbSymbolBypassRevLimit(bool bypass_limit)
 {
     bStrictRevisionLimit = !bypass_limit;
 }
 
 void XbSymbolOutputMessage(xb_output_message mFlag, const char* message)
```
JayFoxRox: nit: mFlag should be renamed to mLevel probably

XbSymbolDatabase.c
RadWolfie: Hmm, should had been `%hu` since version is `unsigned short`.

XbSymbolDatabase.h
RadWolfie:
```
WARNING: OOVPA signature is too high for [5558] D3DDevice_SetTransform!
WARNING: OOVPA signature is too high for [5558] D3DDevice_SetVertexShaderConstantNotInline!
WARNING: Duplicate OOVPA signature found for CDirectSound_SetPosition, 4134 vs 4627!
```

Usually top two lines are tells us either we need to move the OOVPAs to earlier revision or need to strength the pairs in some way.

Whilst last third line can provide early notice plus a chance of showing the signature is too high as well.

XbSymbolDatabase.h
RadWolfie: Xbe's revision to be exact. The scan itself collect a list of each library's revision then determine which one is the highest. Then use the highest revision to end the scan process in order to avoid false positive. Since there are couple/some symbols can only use a offset value which does go up then back down to same value. After all, those symbols do have complete asm codes being identical except for the offset.

XbSymbolDatabase.c
RadWolfie: > The old style of calling it is still valid

Format string as message perform things differently plus bad code can cause an exception without awareness. Plus C standard doesn't allow two same function names with different parameters.

> it's not exactly performance critical either.

Fee free to scatter many `tempBuffer[2048]`and many of snprintf/printf calls around in multiple functions. I'm sure the stacks will love multiple big local variable sizes. 🙄

JayFoxRox: > Format string as message perform things differently plus bad code can cause an exception without awareness.

C doesn't have exceptions.

*(The most dangerous part in your implementation would be a buffer-overflow, which can happen regardless)*

> Plus C standard doesn't allow two same function names with different parameters.

I did not mean an overload: You can change the original function (as you also implied in the second half of your response).

Unless we did `XbSymbolOutputMessage(buffer)` which might allow "%" to be present, this is a non-issue.

This is similar to how you'd always use `printf("Hello World!\n")` instead of `puts("Hello World!\n")`. It's not an issue unless you write dangerous code to begin with.

This is what I meant with "The old style of calling it is still valid".

> Fee free to scatter many tempBuffer[2048]and many of snprintf/printf calls around in multiple functions. I'm sure the stacks will love multiple big local variable sizes.

This function is not recursive. By the time it leaves, the memory is already gone again.

As this is the bottom of the stack, and not initialized, this reduces to a single add/sub instruction - it doesn't cost performance except for a page fault (if you are unlucky), which doesn't matter either because it's not a hot-loop and will itself call all over the place (`output_func`).

> EDIT: Also, we don't need to waste extra CPU cycles for formatting the string if it's not going to be output at all.

I don't think a bit of performance output critical, especially as these functions are only ran once and the result is even cached in most applications.

I'd also assume that compilers optimize the `vsprintf` call. All popular compilers definitely *do* optimize `printf` into `puts` if no `%` is present at compile time; I assume the same happens to `vsprintf` which probably turns into `strcpy` (I did not check though).

XbSymbolDatabase.c
```diff
@@ -202,24 +202,32 @@ void XbSymbolBypassRevLimit(bool bypass_limit)
     bStrictRevisionLimit = !bypass_limit;
 }
 
-void XbSymbolOutputMessage(xb_output_message mFlag, const char* message)
+// Intended for send the message as-is without formating.
+// (To avoid corrupted string when perform a bad coding.
```
JayFoxRox: If this happens, the calling code should be fixed - falling back to a "safer" but possibly erroneous print shouldn't even be an option.

XbSymbolDatabase.c
```diff
@@ -181,9 +187,47 @@ void XbSymbolSetOutputMessage(xb_output_message_t message_func)
...
+    // Plus if pass minimum verbose level to output.
+    if (output_func != NULL && mLevel >= output_verbose_level) {
+        output_func(mLevel, message);
+    }
+}
+
+// Intended to format the message with given extra parameters.
+void XbSymbolOutputMessageFormat(xb_output_message mLevel, const char* format, ...)
+{
+    static char bufferTemp[2048];
```
JayFoxRox: This isn't thread-safe.

Thread extension is *cheap*. I'd suggest keeping this on the stack.

RadWolfie: *Oops*

XbSymbolDatabase.h
JayFoxRox: Okay.

Minor notes which affect the code comment:

> Xbe's revision to be exact.

There's no such field in the XBE headers. I assume you meant assembled version info from the libraries, as explained afterwards.

> The scan itself collect a list of each library's revision

I assume you mean "library build numbers" - this is important because "revision" refers to something else. The Alpha XDKs had a binary versioned with ` 3146.3 ` for example, which is build 3146, but QFE / Revision 3. Also see <https://docs.microsoft.com/en-us/windows/desktop/api/shlwapi/nf-shlwapi-makedllverull>

XbSymbolDatabase.h
```diff
@@ -485,7 +485,8 @@ uint32_t XbSymbolLibrayToFlag(const char* library_name);
 unsigned int XbSymbolDataBaseTestOOVPAs();
 
 /// <summary>
-/// (Debug feature) Set to true will perform full range of OOVPAs registered. Or stop at title's revision detected.
+/// (Debug feature) Set to true will perform full range of OOVPAs registered in current database.
+/// Or stop at xbe's build revision detected.
```
JayFoxRox: *"build version" or "build number"

RadWolfie:
Any further review before merge this pull request?

RadWolfie:
Debug features are working as intended.

PR #61, opened 2019-03-21
Minor improvement

RadWolfie:
* Fixed OV_BYTE bug since only value is acceptable for 1 byte size.
  * However, current codebase isn't using the OV_ENTRY macro.
  * Current codebase will eventually replace default OV entries to macro usage.
* Rename OOVPA_TABLE_COUNT to XBSDB_ARRAY_SIZE

Plus a new macro, **OV_MATCH**, for ability to insert multiple values in one line instead of multiple lines.
image

Image proof of above image expands to:
image

### DIFF
--- a/OOVPA.h
+++ b/OOVPA.h
@@ -218,8 +218,6 @@ typedef struct _OOVPATable
      _9,  _8,  _7,  _6,  _5,  _4,  _3,  _2,  _1,  _0, \
     N, ...) N
 
-// NOTE: A reminder for all of developers, TYPE argument is meant for research documentation.
-// (eventually UNPATCHED will be replace back to PATCH once LLE is fully supportive in the future)
 #define REGISTER_OOVPAS(Symbol, ...) MSVC_EXPAND(_GET_NTH_ARG("ignored", __VA_ARGS__, \
     REGISTER_OOVPA_19, REGISTER_OOVPA_18, REGISTER_OOVPA_17, REGISTER_OOVPA_16, REGISTER_OOVPA_15, \
     REGISTER_OOVPA_14, REGISTER_OOVPA_13, REGISTER_OOVPA_12, REGISTER_OOVPA_11, REGISTER_OOVPA_10, \
--- a/OOVPA.h
+++ b/OOVPA.h
@@ -72,8 +72,20 @@ typedef struct _LOVP
     // Both the Offset and Value are 16-bit to allow for XRefs with a
     // large offset. Value can be safely cast to 8-bit for OOVPA, but must
     // remain 16-bit for XRef entries.
-    unsigned short Offset;
-    unsigned short Value;
+    unsigned short offset;
+    union {
+        // Generic structure TODO: Remove me after replace array to macro for OOVPA database
+        unsigned short value;
+
+        struct {
+            unsigned char unused;
+            unsigned char value;
+        } byte;
+
+        struct {
+            unsigned short index;
+        } xref;
+    };
 } LOVP;
 
 // This XRefZero constant, when set in the OOVPA.XRefCount field,
@@ -96,12 +108,14 @@ typedef struct _LOVP
 
 // Macro used for storing an XRef {Offset, XREF}-Pair.
 #define XREF_ENTRY( Offset, XRef)    \
-    { Offset, XRef }
+    { Offset, .xref.index = XRef }
 
-// UNUSED Macro for storing a normal (non-XRef) {Offset, Value}-Pair
+// Macro for storing a normal (non-XRef) {Offset, Value}-Pair
 // Offsets can go up to 16 bits, values are always one byte (8 bits)
 #define OV_ENTRY(Offset, Value)    \
-    { Offset, Value }
+    { Offset, .value = Value }
+// TODO: Remove above line, then uncomment the line below for easy swap after OOVPA database has been fixed.
+//  { Offset, { .unused = 0, .value = Value } } 
 
 
 // ******************************************************************
--- a/OOVPA.h
+++ b/OOVPA.h
@@ -154,8 +154,8 @@ typedef struct _OOVPATable
 
 // http://en.cppreference.com/w/cpp/iterator/size
 //#include <iterator>
-//#define OOVPA_TABLE_COUNT(x) std::size(x)
-#define OOVPA_TABLE_COUNT(x) (sizeof(x)/sizeof(x[0]))
+//#define XBSDB_ARRAY_SIZE(x) std::size(x)
+#define XBSDB_ARRAY_SIZE(x) (sizeof(x)/sizeof(x[0]))
 
 #define OOVPA_TABLE_ENTRY_FULL(Oovpa, DebugName, Version) \
     { & Oovpa ## _ ## Version.Header, DebugName, Version }
--- a/OOVPA.h
+++ b/OOVPA.h
@@ -41,21 +41,21 @@
 // ******************************************************************
 typedef struct _OOVPA
 {
-    // This OOVPA field (uint08 Count) indicates the number of
+    // This OOVPA field, Count, indicates the number of
     // {Offset, Value}-pairs present in the Lovp array,
     // available after casting this OOVPA to LOOVPA.
     // (This Count INCLUDES optional leading {Offset, XREF_*-enum}-
     // pairs - see comment at XRefCount.)
     unsigned char Count;
 
-    // This OOVPA field (uint08 XRefCount) contains the number of
+    // This OOVPA field, XRefCount, contains the number of
     // {Offset, XREF_*-enum}-pairs that come before all other
     // {Offset, Value}-pairs.
     // (The {Offset, XREF_*-enum}-pairs are INCLUDED in OOVPA.Count)
     // (Also, see comments at XRefZero and XRefOne.)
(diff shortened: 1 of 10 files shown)


OOVPA.h
```diff
@@ -182,14 +209,18 @@ typedef struct _OOVPATable
...
 
-// NOTE: A reminder for all of developers, TYPE argument is meant for research documentation.
-// (eventually UNPATCHED will be replace back to PATCH once LLE is fully supportive in the future)
 #define REGISTER_OOVPAS(Symbol, ...) MSVC_EXPAND(_GET_NTH_ARG("ignored", __VA_ARGS__, \
     REGISTER_OOVPA_19, REGISTER_OOVPA_18, REGISTER_OOVPA_17, REGISTER_OOVPA_16, REGISTER_OOVPA_15, \
     REGISTER_OOVPA_14, REGISTER_OOVPA_13, REGISTER_OOVPA_12, REGISTER_OOVPA_11, REGISTER_OOVPA_10, \
     REGISTER_OOVPA_9, REGISTER_OOVPA_8, REGISTER_OOVPA_7, REGISTER_OOVPA_6, REGISTER_OOVPA_5, \
     REGISTER_OOVPA_4, REGISTER_OOVPA_3, REGISTER_OOVPA_2, REGISTER_OOVPA_1, REGISTER_OOVPA_0)(Symbol, __VA_ARGS__))
 
+#define OV_ENTRIES(Offset, ...) MSVC_EXPAND(_GET_NTH_ARG("ignored", __VA_ARGS__, \
```
JayFoxRox: nit: I think a name like `OV_BYTES` would be more fitting. As the macro is only used in those files, you could choose an even shorter or descriptive name without prefix (my original proposal on XboxDev Discord was `MATCH(0x####, 0x##, 0x##, ...)`).

I also think the trailing comma could be part of the macro, so that it will eventually look like a domain specific language, parsable by trivial non-C compilers.

This can also be seen in my original proposal:

```
        // D3DDevice_GetBackBuffer+0x04 : cmp eax, 0xFFFFFFFF
        MATCH(0x0004,     0x83, 0xF8, 0xFF)

        // D3DDevice_GetBackBuffer+0x0D : jnz +0x07
        MATCH(0x000D,     0x75, 0x07)
```

OOVPA.h
```diff
@@ -139,8 +145,8 @@ typedef struct _OOVPATable
 
 // http://en.cppreference.com/w/cpp/iterator/size
 //#include <iterator>
-//#define OOVPA_TABLE_COUNT(x) std::size(x)
-#define OOVPA_TABLE_COUNT(x) (sizeof(x)/sizeof(x[0]))
+//#define XBSDB_ARRAY_SIZE(x) std::size(x)
+#define XBSDB_ARRAY_SIZE(x) (sizeof(x)/sizeof(x[0]))
```
JayFoxRox: If OOVPA.h is only used internally, I'd suggest removing this prefix (and make it `ARRAY_SIZE` - which is a common name for this).
If it's also used externally, this should be changed, and the macro should probably be removed (it's not used within this file, but is very generic).

OOVPA.h
```diff
@@ -172,6 +178,27 @@ typedef struct _OOVPATable
...
+ #define OV_ENTRY_10(Offset, Value, ...) OV_ENTRY(Offset, Value), MSVC_EXPAND(OV_ENTRY_9(Offset+1, __VA_ARGS__))
+ #define OV_ENTRY_11(Offset, Value, ...) OV_ENTRY(Offset, Value), MSVC_EXPAND(OV_ENTRY_10(Offset+1, __VA_ARGS__))
+ #define OV_ENTRY_12(Offset, Value, ...) OV_ENTRY(Offset, Value), MSVC_EXPAND(OV_ENTRY_11(Offset+1, __VA_ARGS__))
+ #define OV_ENTRY_13(Offset, Value, ...) OV_ENTRY(Offset, Value), MSVC_EXPAND(OV_ENTRY_12(Offset+1, __VA_ARGS__))
+ #define OV_ENTRY_14(Offset, Value, ...) OV_ENTRY(Offset, Value), MSVC_EXPAND(OV_ENTRY_13(Offset+1, __VA_ARGS__))
+ #define OV_ENTRY_15(Offset, Value, ...) OV_ENTRY(Offset, Value), MSVC_EXPAND(OV_ENTRY_14(Offset+1, __VA_ARGS__))
+ #define OV_ENTRY_16(Offset, Value, ...) OV_ENTRY(Offset, Value), MSVC_EXPAND(OV_ENTRY_15(Offset+1, __VA_ARGS__))
+ #define OV_ENTRY_17(Offset, Value, ...) OV_ENTRY(Offset, Value), MSVC_EXPAND(OV_ENTRY_16(Offset+1, __VA_ARGS__))
+ #define OV_ENTRY_18(Offset, Value, ...) OV_ENTRY(Offset, Value), MSVC_EXPAND(OV_ENTRY_17(Offset+1, __VA_ARGS__))
+ #define OV_ENTRY_19(Offset, Value, ...) OV_ENTRY(Offset, Value), MSVC_EXPAND(OV_ENTRY_18(Offset+1, __VA_ARGS__))
```
JayFoxRox: nit: I believe the longest x86 instruction can only be 16 bytes; so 19 bytes might be overkill - not worth changing though.

OOVPA.h
```diff
@@ -98,10 +104,10 @@ typedef struct _LOVP
 #define XREF_ENTRY( Offset, XRef)    \
     { Offset, XRef }
 
-// UNUSED Macro for storing a normal (non-XRef) {Offset, Value}-Pair
+// Macro for storing a normal (non-XRef) {Offset, Value}-Pair
 // Offsets can go up to 16 bits, values are always one byte (8 bits)
```
JayFoxRox: What's the reason for limiting this to 8 bits in the first place?

OOVPA.h
```diff
@@ -73,7 +73,13 @@ typedef struct _LOVP
     // large offset. Value can be safely cast to 8-bit for OOVPA, but must
     // remain 16-bit for XRef entries.
     unsigned short Offset;
-    unsigned short Value;
+    union {
+        unsigned short sValue;
+        struct {
+            unsigned char unused;
+            unsigned char cValue;
```
JayFoxRox: Weird prefix style. You'd normally see `b` for byte (`unsigned char`) in MS style code.

OOVPA.h
```diff
@@ -98,10 +104,10 @@ typedef struct _LOVP
 #define XREF_ENTRY( Offset, XRef)    \
     { Offset, XRef }
 
-// UNUSED Macro for storing a normal (non-XRef) {Offset, Value}-Pair
+// Macro for storing a normal (non-XRef) {Offset, Value}-Pair
 // Offsets can go up to 16 bits, values are always one byte (8 bits)
 #define OV_ENTRY(Offset, Value)    \
```
JayFoxRox: #62 suggests to use `OV_ENTRIES`, but also `OV_ENTRY`. I'd recommend to only use one name, which can handle 1-N arguments.

```
MATCH(0x0000, 0x01)
MATCH(0x0000, 0x01, 0x02)
```

should work; so there's no benefit doing (it will only be confusing and make writing additional parsers harder):

```
MATCH1(0x0000, 0x01)
MATCHn(0x0000, 0x01, 0x02)
```

OOVPA.h
```diff
@@ -98,10 +104,10 @@ typedef struct _LOVP
...
-    { Offset, Value }
+    { Offset, { .unused=0, .cValue=Value } }
```
JayFoxRox: One could use the .unused into a flags field, so it would be possible to do:

```
MATCH(0x0000, 0x01, 0xXX, 0x02, 0x03)
```

which has a wildcard, which is encoded as field that's being ignored (the same could obviously be used for X-Ref references by storing flags bits only in the highest bits; one could still have 14 bits for the X-Ref index, with 2 bits as type [ignored/x-ref/value/<reserved>]).

Anyhow, this needs some more consideration probably

OOVPA.h
RadWolfie: `ARRAY_SIZE` from my understanding aren't included in C standard files nor is included with every compilers. Part of the reason I used the prefix is to avoid macro redefinition plus prevent overridden by some bad macro.

OOVPA.h
RadWolfie: At least it give you ability to add second instruction in same line. Assuming, it's not over 19 bytes.

We had put a comments with 2+ instructions in one comment. Though, not every values are used, some are commented out to reduce the count and is not unique values.

OOVPA.h
JayFoxRox: As macros are only part of that specific compilation unit, you don't have to worry about a prefix.

Worst case is we'll some day have to add `#undef ARRAY_SIZE` or `#ifndef ARRAY_SIZE` - both extremely unlikely as such macros are only used in projects, and not in headers which are exposed to others.

OOVPA.h
RadWolfie: (won't fix as nothing is implemented in this pull request)

OOVPA.h
RadWolfie: Been replaced to OV_MATCH

RadWolfie:
Merging as is.

PR #67, opened 2019-05-04
Fix NXDK support

JayFoxRox:
* nxdk declares `wchar_t` in stddef, so we need to include it.
* Fixes a bug (with warning) introduced in #54 where https://docs.microsoft.com/en-us/cpp/intrinsics/bitscanreverse-bitscanreverse64?view=vs-2019 takes `unsigned long` pointer as first argument.

### DIFF
--- a/XbSymbolDatabase.c
+++ b/XbSymbolDatabase.c
@@ -35,6 +35,7 @@
 #include <stdio.h>
 #include <stdarg.h>
 #include <stdint.h>
+#include <stddef.h>
 
 #ifdef _MSC_VER
 #include <intrin.h>
--- a/XbSymbolDatabase.c
+++ b/XbSymbolDatabase.c
@@ -41,7 +41,7 @@
 #include <intrin.h>
 static inline uint32_t BitScanReverse(uint32_t value)
 {
-    uint32_t index;
+    unsigned long index;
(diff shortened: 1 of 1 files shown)


RadWolfie:
The problem is the return has to be 4 bytes (uint32_t). Since some compilers are using 8 bytes for long type. Plus the pointer is just a "pointer".

Sources:
* https://docs.microsoft.com/en-us/cpp/cpp/data-type-ranges (long == 4 bytes)
* https://docs.oracle.com/cd/E19253-01/817-6223/chp-typeopexpr-2/index.html (long == 8 bytes)
* https://stackoverflow.com/a/39207744 (long == 8 bytes, Cygwin's doc)

Except `__builtin_clz` is using `unsigned int`. My opinion is we may need to make a typedef to use either int or long.

---

Original comment

Base on further investigation in the source code. I do not see the need to keep as `uint32_t` type for `derive_indices` and `derive_index`. Although, GetXRefEntry would need to replace int to unsigned long.

https://github.com/Cxbx-Reloaded/XbSymbolDatabase/blob/3f14046d9aac8ec9b6dc0c3a5e0208af5faf61b5/XbSymbolDatabase.c#L411-L459

@JayFoxRox, Would you like me to update the above advice changes to use `unsigned long` type?

PR #68, opened 2019-05-17
Relocate JVS into Separate Section

RadWolfie:
Separate JVS symbols from XAPI library into its own library. Plus nothing had been changed in OOVPAs except for replace default array to OV_MATCH macro.

The documentation in JVS notes are used by @LukeUsher record:
image

### DIFF
--- a/XbSymbolDatabase.c
+++ b/XbSymbolDatabase.c
@@ -74,6 +74,7 @@ static inline uint32_t BitScanReverse(uint32_t value)
 #include "OOVPADatabase/D3D8_OOVPA.inl"
 #include "OOVPADatabase/D3D8LTCG_OOVPA.inl"
 #include "OOVPADatabase/DSound_OOVPA.inl"
+#include "OOVPADatabase/JVS_OOVPA.inl"
 #include "OOVPADatabase/XGraphic_OOVPA.inl"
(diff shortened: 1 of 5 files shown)


XbSymbolDatabase.c
```diff
@@ -120,6 +121,9 @@ SymbolDatabaseList SymbolDBList[] = {
     // DSOUNDH is just meant to define hot fix, there is no separate section
     //{ XbSymbolLib_DSOUNDH,{ Sec_DSOUND }, &DSound_OOVPAV2, DSound_OOVPA_COUNT },
 
+    // Only used in Chihiro titles and dashboard
```
JayFoxRox: Make clear this applies to Chihiro dashboard, not Xbox dashboard: "and Chihiro dashboard".

(dashboard is also a wrong name here, so ideally replace that with the appropriate name [segaboot?])

RadWolfie: Chihiro applications is better to use. Plus updated the commit to use the reference.

RadWolfie:
**NOTE:** Require pull request #69 to merge first before add another extra check.

RadWolfie:
Now JVS library is able to force scan base on xbe type check unless XbSymbolLib_JVS flag is excluded.

@LukeUsher, I'm ready for approval.

PR #69, opened 2019-05-19
Fix Specific Library Scan in XbSymbolScan API

RadWolfie:
Title says it all.

Apparently this bug has exist since the introduction of `XbSymbolRegisterLibrary` API. `g_library_flag` was used in `XbSymbolSectionScan` until it was disabled due to not being functional ever since.


RadWolfie:
Base on custom XbSymbolUnitTest, I can confirm specific library and all libraries are working.

PR #70, opened 2019-05-20
Detect xbe Type

RadWolfie:
Plus output debug message

### DIFF
--- a/XbSymbolDatabase.c
+++ b/XbSymbolDatabase.c
@@ -174,6 +174,24 @@ xb_output_message output_verbose_level = XB_OUTPUT_MESSAGE_INFO;
 bool bOneTimeScan = true;
 bool bScanFirstDetect = false;
 
+// ported from Dxbx's XbeExplorer
+static xb_xbe_type GetXbeType(const xbe_header *pXbeHeader)
+{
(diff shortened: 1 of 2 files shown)


JayFoxRox:
Not sure if this is a useful feature for this library.

The caller still needs its own XBE parser anyway and checking the type is easy or assumptions can be made. Which kind of libraries are being scanned is also controlled by caller, so why would XbSymbolDatabase need to diverge this info itself?

RadWolfie:
Having this function will skip the JVS library scan if the title is retail.

JayFoxRox:
Seems unnecessary when the caller can do `XbSymbolLib_ALL & ~XbSymbolLib_JVSLIB` on their own.

Most users will know wether the XBE is chihiro or not (because hardware is original Xbox, or it's an emultor that's emulating extra-components to begin with). Even if the user doesn't know (debugger?) then it can easily check on its own (and likely will do that when decoding the entry-point).

RadWolfie:
XbSymbolScan function only perform a check against the library version array. Xbox applications with XDK `4039` build did not include `DSOUND` library version except for the section array. Which require a manual force check if DSOUND library do exist then scan for it.

JVS library is not included in the library version array. Hence the reason for this pull request.

PR #81, opened 2019-07-14
D3D8LTCG Database Optimize

RadWolfie:
Since all library flags are masks, we can simply add `XbSymbolLib_D3D8LTCG` to D3D8's database. Doing so will reduce need to sync the changes from D3D8 to D3DLTCG database. There should not be any drawback since the performance should be the same as before.

LTCG applications has few new symbols detected which are:
- D3DDevice_CreateCubeTexture
- D3DDevice_CreateIndexBuffer
- D3DDevice_CreateTexture
- D3DDevice_CreateVertexBuffer
- D3DDevice_GetRenderTarget
- D3DDevice_SetVertexShaderConstant
- D3D_SetPushBufferSize

And these symbols are confirmed positive finding. Although, I did include xref to some of the symbols to avoid false positives as possible for the future discovery.

Plus also removed "V2" from `_OOVPA` prefix.


XbSymbolDatabase.c
```diff
@@ -1386,11 +1386,14 @@ bool XbLibraryScan(custom_scan_func_t custom_scan_func,
...
                             continue;
                         }
-                        return scan_ret;
                     }
                 }
             }
-            break;
+            // We're not using force break here since we are combining LTCG flag to std flag's oovpa database such as D3D8.
+            // Plus break out of the loop when custom manual scan is done.
+            if (scan_ret) {
```
PatrickvL: The break could still be enabled, if `LibraryFlag` contains only one bit set.

This could be checked like : ``(LibraryFlag & LibraryFlag-1) == 0)`

XbSymbolDatabase.c
```diff
@@ -1512,13 +1515,14 @@ bool XbSymbolScan(const void* xb_header_addr,
...
-                                        XbSymbolScanOOVPA(SymbolDBList[d2].OovpaTable, SymbolDBList[d2].OovpaTableCount, LibraryStr, SymbolDBList[d2].LibSec.library,
+                                        XbSymbolScanOOVPA(SymbolDBList[d2].OovpaTable, SymbolDBList[d2].OovpaTableCount, LibraryStr, LibraryFlag,
                                             pSectionScan, BuildVersion, register_func, xb_start_virt_addr);
                                         break;
                                     }
                                 }
                             }
-                            break;
+                            // We're not using break here since we are combining LTCG flag to std flag's oovpa database such as D3D8.
+                            //break;
```
PatrickvL: Here, same remark as above (enabling `break` when only one bit is set in `LibraryFlag`)

XbSymbolDatabase.c
RadWolfie: That is a good idea, though... it would save a few cpu cycles. But I'll go ahead re-update to use your idea design.

RadWolfie: Hmm... I realize it will not work because `LibraryFlag` will always have one bit flag set. I'll investigate another possible solution.

RadWolfie: Found an alternative solution. `(SymbolDBList[d2].LibSec.library & ~LibraryFlag) == 0` will check if there are other value sets in the database's library flags.

Plus is confirmed working by do a debug test for each value inspection. And did not break the symbol cache file comparison.

XbSymbolDatabase.c
```diff
@@ -1383,14 +1387,17 @@ bool XbLibraryScan(custom_scan_func_t custom_scan_func,
...
+                            // let's return true here instead of waste the loops for nothing.
+                            return scan_ret;
                         }
-                        return scan_ret;
                     }
                 }
             }
-            break;
+            // Use the break if there are 2+ bit flags set such as include LTCG flag in std flag's oovpa database like D3D8.
+            if ((SymbolDBList[d2].LibSec.library & ~LibraryFlag) == 0) {
```
RadWolfie: However, if we do this way... and the database aren't in order such as LTCG database could be at the top and the std database below it. It will only stop at the LTCG database.

XbSymbolDatabase.c
```diff
@@ -106,55 +106,57 @@ typedef const struct _SymbolDatabaseList {
 
 SymbolDatabaseList SymbolDBList[] = {
     // Support inline functions in .text section
-    { XbSymbolLib_D3D8,{ Sec_text, Sec_D3D, Sec_FLASHROM }, D3D8_OOVPAV2, D3D8_OOVPA_COUNT },
+    { XbSymbolLib_D3D8 | XbSymbolLib_D3D8LTCG,{ Sec_text, Sec_D3D, Sec_FLASHROM }, D3D8_OOVPA, D3D8_OOVPA_COUNT },
 
-    // Cannot support LTCG in HLE
-    { XbSymbolLib_D3D8LTCG,{ Sec_D3D }, D3D8LTCG_OOVPAV2, D3D8LTCG_OOVPA_COUNT },
+    // NOTE: May will not have full support to scan every symbols.
```
PatrickvL: `May not have` is a bit vague an expression. Perhaps say:
`This libraries contains LCTG-specific OOVPA not present in the main D3D library, which this is an extension of.`
or something along those lines.

RadWolfie:
```
    // NOTE: LTCG is a compiler optimization extension which may not be found in standard ___ library.
    //       Not every LTCG symbols can be reliable to create signatures depending on the situation.
```

After made a 3rd draft, I believe should be enough to clarify the situation may occur any time in the future. If not, I'm willing to hear what the note should say.

RadWolfie: I had drop the notes about the LTCG comment from this pull request. It should be documented in a Wiki than repeating same notes in the source code.

PR #83, opened 2019-07-17
Implement Specific Detected Selection Scan for OOVPA

RadWolfie:
With specific detected selection scan for OOVPA, it is become possible to have 2+ identical OOVPAs with a different symbol name.

Using this method will also reduce false warnings and errors as well.

### DIFF
--- a/OOVPA.h
+++ b/OOVPA.h
@@ -60,6 +60,10 @@ typedef struct _OOVPA
     // no XREF_* enum defined for this OOVPA.
     unsigned short XRefSaveIndex;
 
+    unsigned char DetectSelect;
+
+    unsigned char Padding[3];
+
     // Define LOVP here to reduce type definition complexity.
     // (Otherwise, if defined in the template classes, that would mean
     // that for each template instance, the type is redefined. Let's
@@ -88,18 +92,22 @@ typedef struct _LOVP
     };
 } LOVP;
 
+// Set variable's padding to 0.
+#define VARPADSET 0
+
 // This XRefZero constant, when set in the OOVPA.XRefCount field,
 // indicates there are no {offset, XREF_*-enum} present in the OOVPA.
-#define XRefZero 0x00
+#define XRefZero    0
 
 // This XRefOne constant, when set in the OOVPA.XRefCount field,
 // indicates the OOVPA contains one (1) {offset, XREF_* enum} pair.
-#define XRefOne 0x01
+#define XRefOne     1
 
 // Sometimes, there can be more than one {Offset, XREF_*-enum}
 // pair at the start of the OOVPA's.
-#define XRefTwo 0x02
-#define XRefThree 0x03
+#define XRefTwo     2
+#define XRefThree   3
+#define XRefFour    4
 
 // This XRefNoSaveIndex constant, when set in the OOVPA.XRefSaveIndex
 // field, functions as a marker indicating there's no XREF_* enum
@@ -118,6 +126,12 @@ typedef struct _LOVP
 // TODO: Remove above line, then uncomment the line below for easy swap after OOVPA database has been fixed.
 //  { Offset, { .unused = 0, .value = Value } } 
 
+// Macro for determine how the OOVPA scan will process for detection.
+#define DetectDefault 0
+#define DetectFirst   1
+#define DetectSecond  2
+#define DetectThird   3
+#define DetectFourth  4
 
 // ******************************************************************
 // * Large Optimized (Offset,Value)-Pair Array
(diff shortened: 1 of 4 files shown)


RadWolfie:
Here's a functional example to able know which detected address to choose from.
```
DEBUG  : Skipped symbol detection found for D3DDevice__ManualFindEventHandleGeneric_3911 (3911), 0x0020e151
----------------------------------------------------
DEBUG  : Skipped symbol detection found for CDirectSoundStream_SetEG (3911), 0x0024ec1d
Symbol Detected: (b3911) 0x0024e41b -> CDirectSoundStream_SetEG
DEBUG  : Skipped symbol detection found for CDirectSoundStream_SetFilter (3911), 0x0024ec31
Symbol Detected: (b3911) 0x0024e42f -> CDirectSoundStream_SetFilter
...
----------------------------------------------------
DEBUG  : Skipped symbol detection found for IDirectSoundStream_SetEG (3911), 0x0024e41b
Symbol Detected: (b3911) 0x0024ec1d -> IDirectSoundStream_SetEG
DEBUG  : Skipped symbol detection found for IDirectSoundStream_SetLFO (3911), 0x0024e407
Symbol Detected: (b3911) 0x0024ec09 -> IDirectSoundStream_SetLFO
DEBUG  : Skipped symbol detection found for IDirectSoundStream_SetFilter (3911), 0x0024e42f
Symbol Detected: (b3911) 0x0024ec31 -> IDirectSoundStream_SetFilter
...
```

RadWolfie:
Updated the comments for better details.

PatrickvL:
How can we ever reliably predict the order in which duplicate functions appear in every possible xbe? This solution, although an admirable attempt, doesn't seem like a solid solution.

For functions that are part of classes (like the sound classes), it'd be much better if each class vmt could be derived and used as xref source, for example.

RadWolfie:
Unfortunately, there is no VMT for the sound functions I included in this pull request. Beside, it doesn't really matter which order for C____ and I____ when they're both exactly the same. Without generated pdb file from the compiler, there's absolutely no chance to able determine which symbol function is which for exactly the same asm functions. If the C___ and I___ are not at the top high level functions, then there's a chance we can fix it. Except they're at the top high level functions.

Plus the generic `D3DDevice__ManualFindEventHandleGeneric_3911` scan is exactly identical in several functions and is unique signature, we only need the first one.

RadWolfie:
FYI, this pull request is currently blocker for design API v3.

PatrickvL:
This pull request lacks a technical explanation, but looking at the code changes, it seems a feature was added that allows to skip a hit a configurable number of times while scanning through an xbe.
This implies the number of skips depends on the number of times, and thus the order in which, identical pattern matches appear in an xbe.
This looks to me like an arbitrary solution, as the order in which functions appear (repeatedly) in an xbe can differ, for various reasons: linker version and flags, xdk version, game code usage of said functions (or not), etc.

Therefore, this needs either more explanation or additional discussion and then we'll see how to continue

RadWolfie:
> This looks to me like an arbitrary solution, as the order in which functions appear (repeatedly) in an xbe can differ, for various reasons: linker version and flags, xdk version, game code usage of said functions (or not), etc.

Right, the XDK library and game code has a mix usage. For DirectSound(Buffer|Stream) class references:
* `CMcpxVoiceClient_SetEG` (lowest level reference)
  * `CDirectSoundVoice_SetEG`
    * `IDirectSoundBuffer_SetEG` (doesn't have vmt and is a direct call.)
    * `CDirectSoundStream_SetEG` (doesn't have vmt and is a direct call.)
    * `IDirectSoundStream_SetEG` (doesn't have vmt and is a direct call.)

Since both `CDirectSoundStream_SetEG` and `IDirectSoundStream_SetEG` doesn't have vmt at the very top level references. As you can see at the lines (+ etc) below, it is reference differently and doesn't have the above CDirectSoundStream_(SetEG/xxxx) functions in it.
https://github.com/Cxbx-Reloaded/XbSymbolDatabase/blob/e9e976850eefc670a6142ea590bbcb8f0026b3c4/XbSymbolDatabase.c#L1347-L1356

As for D3D8, in LTCG, there are actually 2 separate `D3DDevice_MakeSpace` functions! 😄 I had done the analysis has confirm there are two separate "duplicate" `D3DDevice_MakeSpace` functions. This make the single xref broken, which of course will need another update to use two different references in the same signature data. Plus it will create `D3DDevice_MakeSpace` *and* `D3DDevice_MakeSpace2` signatures with this pull request feature requirement. Or don't use `D3DDevice_MakeSpace` reference in existing codebase at all.

Finally, the reason for unique `D3DDevice__ManualFindEventHandleGeneric` signature since the D3D8's standard and LTCG library does change a lot over time. The unique signature method is easier to implement manually without the need to create multiple signatures hopefully one of two or three symbol signatures find itself. (I had issues in the past with just one symbol signature doesn't always exist in the xbe file.) And we don't need the warning/error message about this signature.

The arbitrary solution is the only solution I can discover to resolve the issue even though we don't know which C or I of DirectSoundStream is actually used and both functions, like `SetEG`, process exactly the same way.

PR #89, opened 2019-08-17
Merge XONLINE and XNET into Two Databases

RadWolfie:
* migrate xonlines to xonline filter list
* migrate xnet(s|n) and xonline(s) to xnet filter list
* removed XNET symbols from XONLINE list

Tested with Xbox Live xbe file which has XONLINES library, contain XONLINE and XNET sections, with no issues.

With a bonus, the errors has been reduce from 522 to 154.


RadWolfie:
FYI, all this does is add the library flags to the "filter" list for existing database. Since there is no need for duplicate database entries for the same database per library flag.

PR #90, opened 2019-08-21
Add symbol for D3D_DestroyResource

### LINKED ISSUE #82: OOVPA Symbol Request: DestroyResource
DestroyResource in D3D8.LIB is a function that is called internally by the XDK when a resource needs to be freed.

I recently came across a scenario in Cxbx-R where a resource could be erronously freed under certain circumstances. The solution for this issue would require hooking DestroyResource, but currently no OOVPA exists for this.

Requirement for https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1665

LukeUsher:
This pattern works for all known XDK versions, from 3911-5933

Fix #82

### DIFF
--- a/OOVPADatabase/D3D8/3911.inl
+++ b/OOVPADatabase/D3D8/3911.inl
@@ -418,6 +418,35 @@ OOVPA_NO_XREF(D3DDevice_GetGammaRamp, 3911, 14)
 OOVPA_END;
 
 // ******************************************************************
+// * D3D_DestroyResource
+// ******************************************************************
+// Generic OOVPA as of 3911 and newer.
+OOVPA_NO_XREF(D3D_DestroyResource, 3911, 38)
+    // push edi 
+    // push eda
+    // mov edi [esp + $0C]
+    // mov eax, [edi]
+    // mov esi, eax
+    OV_MATCH(0x00, 0x56, 0x57, 0x8B, 0x7C, 0x24, 0x0C, 0x8B, 0x07, 0x8B, 0xF0),
+
+    // and esi, $70000
+    // cmp esi, $50000
+    OV_MATCH(0x0A, 0x81, 0xE6, 0x00, 0x00, 0x07, 0x00, 0x81, 0xFE, 0x00, 0x00, 0x05, 0x00),
+    
+    // jne eip + $04
+    // test eax, eax
+    // jne eip + $06
+    // push edi
+    OV_MATCH(0x16, 0x75, 0x04, 0x85, 0xC0, 0x79, 0x06, 0x57),
+
+    // cmp esi, $50000
+    OV_MATCH(0x28, 0x81, 0xFE, 0x00, 0x00, 0x05, 0x00),
+
+    // mov eax, [edi + $04]
+    OV_MATCH(0x35, 0x8B, 0x47, 0x04),
+OOVPA_END;
+
+// ******************************************************************
 // * D3D_SetPushBufferSize
 // ******************************************************************
 // Generic OOVPA as of 3911 and newer.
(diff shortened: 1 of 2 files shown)


RadWolfie:
I do not approve this pull request for the main reason, 38 OVs is too much. We do not have OVs over 20s more or less. I'm sure there is better optimization than this.

PatrickvL:
The more entries, the less chance of false positives, a few more bytes then average doesn't really hurt, and perhaps for some OOVPA, a relatively large amount of offset+value pairs is necessary to make it unique. There's no rule that I'm aware of (yet) that we must minimize the size of OOVPA?

RadWolfie:
The more entries, the longer the scan process will take. 😉 I'll make my review to demonstrate what's not really needed soon.

We do have minimal requirement of 10 - 12 (maybe 8? base in the codebase).

OOVPADatabase/D3D8/3911.inl
```diff
@@ -417,6 +417,35 @@ OOVPA_NO_XREF(D3DDevice_GetGammaRamp, 3911, 14)
...
+    // push edi 
+    // push eda
+    // mov edi [esp + $0C]
+    // mov eax, [edi]
+    // mov esi, eax
+    OV_MATCH(0x00, 0x56, 0x57, 0x8B, 0x7C, 0x24, 0x0C, 0x8B, 0x07, 0x8B, 0xF0),
+
+    // and esi, $70000
+    // cmp esi, $50000
+    OV_MATCH(0x0A, 0x81, 0xE6, 0x00, 0x00, 0x07, 0x00, 0x81, 0xFE, 0x00, 0x00, 0x05, 0x00),
```
RadWolfie: Both of these opcodes are unique.

OOVPADatabase/D3D8/3911.inl
```diff
@@ -417,6 +417,35 @@ OOVPA_NO_XREF(D3DDevice_GetGammaRamp, 3911, 14)
...
+    
+    // jne eip + $04
+    // test eax, eax
+    // jne eip + $06
+    // push edi
+    OV_MATCH(0x16, 0x75, 0x04, 0x85, 0xC0, 0x79, 0x06, 0x57),
```
RadWolfie: This is not necessary since they're not unique in any way.

OOVPADatabase/D3D8/3911.inl
```diff
@@ -417,6 +417,35 @@ OOVPA_NO_XREF(D3DDevice_GetGammaRamp, 3911, 14)
...
+
+    // cmp esi, $50000
+    OV_MATCH(0x28, 0x81, 0xFE, 0x00, 0x00, 0x05, 0x00),
```
RadWolfie: Again, is also unique. Though there is no and opcode before this which is a good thing.

OOVPADatabase/D3D8/3911.inl
```diff
@@ -417,6 +417,35 @@ OOVPA_NO_XREF(D3DDevice_GetGammaRamp, 3911, 14)
...
+
+    // mov eax, [edi + $04]
+    OV_MATCH(0x35, 0x8B, 0x47, 0x04),
```
RadWolfie: This is not necessary since it's not unique in any way.

OOVPADatabase/D3D8/3911.inl
```diff
@@ -417,6 +417,35 @@ OOVPA_NO_XREF(D3DDevice_GetGammaRamp, 3911, 14)
         { 0x27, 0xC2 },
 OOVPA_END;
 
+// ******************************************************************
+// * D3D_DestroyResource
+// ******************************************************************
+// Generic OOVPA as of 3911 and newer.
+OOVPA_NO_XREF(D3D_DestroyResource, 3911, 38)
+    // push edi
```
RadWolfie: 0x56 opcode is `push esi` and the below is `push edi` for 0x57 opcode.

PR #94, opened 2019-08-25
Design API v3

### LINKED ISSUE #10: TASK: Design API
To be able to use this XbSymbolDatabase in Cxbx-Reloaded (and possible other projects, like via an IDA Pro plugin), we'll (obviously) be needing an API to access it's functionality.

Things that this library should be able to do :
* query the library for it's version & capabilities
* register symbol-table database(s) [ this might not be needed until #8 is decided upon ]
* start a scanning session, returning a handle
* scan a memory-range for symbols (internally adding the detected symbols to the session handle)
* retrieve the detected symbols in a session
* closing a scanning sessions, freeing all allocated resources

This API must be implemented with thread-safety in mind.
API signatures must be designed such, that future updates to the library could be added without breaking the interface.

The API that will be doing the actual scanning, must be able to be updated in the future, so it might be wise to give it only one argument : a structure pointer that passes in all arguments.
This way, the structure can be changed in the future, without breaking the BAPI.

Things that should go into the scanning struct:
* a struct version
* scanning flags (or a reserved DWORD to put those in once we need them)
* an indication of the range / ranges that should be scanned
  * per range: it's size in bytes, a pointer to it's data, and it's base-address in virtual-memory

Please discuss further below

RadWolfie:
Design API v3.0

Closes #10

New features: (see XbSymbolDatabase.h file for functions' details)
* XbSymbolDatabase_GenerateLibraryFilter
* XbSymbolDatabase_GenerateSectionFilter
* XbSymbolDatabase_CreateXbSymbolContext
* XbSymbolContext_ScanLibrary (design for multi-thread safe support)
   * If there is a dependency on another library's symbols. Such as XACTENG library do depend on DSOUND's symbols. Then second or later libraries will not perform the scan until the library is done scanning in the active thread.
* XbSymbolContext_ScanAllLibraryFilter (design for single-threading only)
   * May will change to support multi-thread usage internally.
* XbSymbolContext_RegisterXRefs (currently is a stub function)
  * Can be resolved by #39
* XbSymbolContext_Release

Currently, the design is backward compatibility. Until final stage will move the old APIs into new APIs.

### DIFF
--- a/XbSymbolDatabase.c
+++ b/XbSymbolDatabase.c
@@ -36,6 +36,11 @@
 #include <stdarg.h>
 #include <stdint.h>
 #include <stddef.h>
+// TODO: Waiting on Visual Studio to include C11 support for multi-thread safe.
+#ifndef _MSC_VER
+#define MULTI_THREAD_SAFE true
+#include <threads.h>
+#endif
 
 #ifdef _MSC_VER
 #include <intrin.h>
@@ -84,23 +89,60 @@ static inline uint32_t BitScanReverse(uint32_t value)
 
 typedef uint8_t* memptr_t;
 
-typedef bool (*custom_scan_func_t)(uint32_t LibraryFlag,
-                                   const xbe_section_header* pSectionHeader,
-                                   unsigned short BuildVersion,
-                                   const char* LibraryStr,
-                                   xb_symbol_register_t register_func,
-                                   memptr_t xb_start_virt_addr);
+typedef struct _OutputHandler {
+    xb_output_message   verbose_level;
+    xb_output_message_t func;
+} OutputHandler;
+
+// Library Type is a requirement to prevent another thread
+// doing the same scan process.
+typedef enum _eLibraryType {
+    LT_D3D = 0,
+    LT_AUDIO,
+    LT_JVS,
+    LT_XAPI,
+    LT_GRAPHIC,
+    LT_NETWORK,
+    LT_UNKNOWN,
+    LT_COUNT=LT_UNKNOWN
+} eLibraryType;
+
+typedef enum _eScanStage {
+    SS_NONE = 0,
+    SS_1_MANUAL,
+    SS_2_SCAN_LIBS,
+} eScanStage;
+
+typedef struct _iXbSymbolContext {
+    bool                    strict_build_version_limit;
+    bool                    one_time_scan;
+    bool                    scan_first_detect;
+    uint32_t                library_filter;
+    xbaddr                  xref_database[XREF_COUNT];
+    OutputHandler           output;
+    xb_symbol_register_t    register_func;
+    XbSDBLibraryHeader      library_input;
+    XbSDBSectionHeader      section_input;
+    bool                    library_active[LT_COUNT];
+    eScanStage              scan_stage;
+#ifdef MULTI_THREAD_SAFE
+    mtx_t                   mutex;
+#endif
+} iXbSymbolContext;
+
+typedef bool (*custom_scan_func_t)(iXbSymbolContext* pContext,
+                                   const XbSDBLibrary* pLibrary,
+                                   const XbSDBSection* pSection);
 
 typedef const struct _PairScanLibSec {
     uint32_t library;
-    const char *section[PAIRSCANSEC_MAX];
+    const char* section[PAIRSCANSEC_MAX];
 } PairScanLibSec;
 
 typedef const struct _SymbolDatabaseList {
+    PairScanLibSec  LibSec;
 
-    PairScanLibSec LibSec;
-
-    OOVPATable     *OovpaTable;
+    OOVPATable*     OovpaTable;
     unsigned int    OovpaTableCount;
 } SymbolDatabaseList;
 
@@ -160,24 +202,59 @@ SymbolDatabaseList SymbolDBList[] = {
 // ******************************************************************
 const unsigned int SymbolDBListCount = XBSDB_ARRAY_SIZE(SymbolDBList);
 
-// ******************************************************************
-// * XRefDataBase
-// ******************************************************************
-xbaddr XRefDataBase[XREF_COUNT] = { 0 }; // Reset and populated by EmuHLEIntercept
+const char SectionList[][8] = {
+    Sec_text,
+    Sec_FLASHROM,
+    Sec_rdata,
+    Sec_D3D,
+    Sec_D3DX,
+    Sec_DSOUND,
+    Sec_XACTENG,
+    Sec_XPP,
+    Sec_XGRPH,
+    Sec_XONLINE,
+    Sec_XNET
+};
 
-bool bXRefFirstPass; // For search speed optimization, set in XbSymbolScan, read in XbSymbolLocateFunction
-unsigned int UnResolvedXRefs = XREF_COUNT;
-bool bStrictBuildVersionLimit = true;
+const unsigned int SectionListTotal = XBSDB_ARRAY_SIZE(SectionList);
+
 #if _DEBUG
-xb_output_message output_verbose_level = XB_OUTPUT_MESSAGE_DEBUG;
+xb_output_message g_output_verbose_level = XB_OUTPUT_MESSAGE_DEBUG;
 #else
-xb_output_message output_verbose_level = XB_OUTPUT_MESSAGE_INFO;
+xb_output_message g_output_verbose_level = XB_OUTPUT_MESSAGE_INFO;
 #endif
-bool bOneTimeScan = true;
-bool bScanFirstDetect = false;
+
+// Intended for send the message as-is without formating.
+// (To avoid corrupted string when perform a bad coding.
+//  Plus certain message need special character without format.)
+void output_message(OutputHandler* pOutput, xb_output_message mLevel, const char* message)
+{
+    // Check if output function is set.
+    // Plus if pass minimum verbose level to output.
+    if (pOutput->func != NULL && mLevel >= pOutput->verbose_level) {
+        pOutput->func(mLevel, message);
+    }
+}
+
+// Intended to format the message with given extra parameters.
+void output_message_format(OutputHandler* pOutput, xb_output_message mLevel, const char* format, ...)
+{
+    char bufferTemp[2048];
+
+    // Check if output function is set.
+    // Plus if pass minimum verbose level to output.
+    if (pOutput->func != NULL && mLevel >= pOutput->verbose_level) {
+        va_list args;
+        va_start(args, format);
+        (void)vsprintf(bufferTemp, format, args);
+        va_end(args);
+
+        pOutput->func(mLevel, bufferTemp);
+    }
+}
 
 // ported from Dxbx's XbeExplorer
-static xb_xbe_type GetXbeType(const xbe_header *pXbeHeader)
+static xb_xbe_type GetXbeType(const xbe_header* pXbeHeader)
 {
     // Detect if the XBE is for Chihiro (Untested!) :
     // This is based on https://github.com/radare/radare2/blob/7ffe2599a192bf5b9333560345f80dd97f096277/libr/bin/p/bin_xbe.c#L29
@@ -195,83 +272,134 @@ static xb_xbe_type GetXbeType(const xbe_header *pXbeHeader)
     return XB_XBE_TYPE_RETAIL;
 }
 
+static bool iXbSymbolContext_Lock(iXbSymbolContext* pContext)
+{
+#ifdef MULTI_THREAD_SAFE
+    // Lock to this thread only during the scan process until the scan is done.
+    int mtxStatus = mtx_lock(&pContext->mutex);
+    if (mtxStatus != thrd_success) {
+        output_message_format(&pContext->output, XB_OUTPUT_MESSAGE_ERROR, "Unable to lock mutex: %d", mtxStatus);
+        return false;
+    }
+#endif
+
+    return true;
+}
+
+static void iXbSymbolContext_Unlock(iXbSymbolContext* pContext)
+{
+#ifdef MULTI_THREAD_SAFE
+    // Lock to this thread only during the scan process until the scan is done.
+    (void)mtx_unlock(&pContext->mutex);
+#endif
+}
+
+static bool iXbSymbolContext_AllowSetParameter(iXbSymbolContext* pContext)
+{
+    bool bRet;
+
+    if (!iXbSymbolContext_Lock(pContext)) {
+        return false;
+    }
+
+    bRet = (pContext->scan_stage == SS_NONE);
+
+    if (!bRet) {
+        output_message(&pContext->output, XB_OUTPUT_MESSAGE_ERROR, "Cannot set parameter value after first and during scan call.");
+    }
+
+    iXbSymbolContext_Unlock(pContext);
+
+    return bRet;
+}
+
+static bool iXbSymbolContext_AllowScanLibrary(iXbSymbolContext* pContext)
+{
+    bool bRet;
+
+    if (!iXbSymbolContext_Lock(pContext)) {
+        return false;
+    }
+
+    bRet = (pContext->scan_stage == SS_2_SCAN_LIBS);
+
+    if (!bRet) {
+        output_message(&pContext->output, XB_OUTPUT_MESSAGE_ERROR, "XbSymbolContext_ScanManual must be call first before scan for library's symbols.");
+    }
+
+    iXbSymbolContext_Unlock(pContext);
+
+    return bRet;
+}
+
 // ******************************************************************
 // * API functions to use with other projects.
 // ******************************************************************
 
-uint32_t g_library_flag = 0;
-bool XbSymbolRegisterLibrary(uint32_t library_flag) {
+bool XbSymbolContext_RegisterLibrary(XbSymbolContextHandle pHandle, uint32_t library_filter)
+{
+    iXbSymbolContext* pContext = (iXbSymbolContext*)pHandle;
 
-    // Check to make sure all flags are acceptable before set.
-    if ((library_flag & ~XbSymbolLib_ALL) > 0) {
-        return 0;
+    if (!iXbSymbolContext_AllowSetParameter(pContext)) {
+        return false;
     }
 
-    g_library_flag = library_flag;
-    return 1;
+    // Check to make sure all flags are acceptable before set.
+    if ((library_filter & ~XbSymbolLib_ALL) > 0) {
+        return false;
+    }
+
+    pContext->library_filter = library_filter;
+    return true;
 }
 
-xb_output_message_t output_func = NULL;
-void XbSymbolSetOutputMessage(xb_output_message_t message_func)
+xb_output_message_t g_output_func = NULL;
+void XbSymbolDatabase_SetOutputMessage(xb_output_message_t message_func)
 {
-    output_func = message_func;
+    g_output_func = message_func;
 }
 
-bool XbSymbolSetOutputVerbosity(xb_output_message verbose_level)
+bool XbSymbolDatabase_SetOutputVerbosity(xb_output_message verbose_level)
 {
     if (verbose_level < XB_OUTPUT_MESSAGE_MAX) {
-        output_verbose_level = verbose_level;
+        g_output_verbose_level = verbose_level;
         return true;
     }
     // Return false only when requested level is invalid.
     return false;
 }
 
-void XbSymbolBypassBuildVersionLimit(bool bypass_limit)
+void XbSymbolContext_SetBypassBuildVersionLimit(XbSymbolContextHandle pHandle, bool bypass_limit)
 {
-    bStrictBuildVersionLimit = !bypass_limit;
-}
+    iXbSymbolContext* pContext = (iXbSymbolContext*)pHandle;
 
-void XbSymbolContinuousSigScan(bool enable)
-{
-    bOneTimeScan = !enable;
-}
-
-void XbSymbolFirstDetectAddressOnly(bool enable)
-{
-    bScanFirstDetect = enable;
-}
-
-// Intended for send the message as-is without formating.
-// (To avoid corrupted string when perform a bad coding.
-//  Plus certain message need special character without format.)
-void XbSymbolOutputMessage(xb_output_message mLevel, const char* message)
-{
-    // Check if output function is set.
-    // Plus if pass minimum verbose level to output.
-    if (output_func != NULL && mLevel >= output_verbose_level) {
-        output_func(mLevel, message);
+    if (!iXbSymbolContext_AllowSetParameter(pContext)) {
+        pContext->strict_build_version_limit = !bypass_limit;
     }
+
 }
 
-// Intended to format the message with given extra parameters.
-void XbSymbolOutputMessageFormat(xb_output_message mLevel, const char* format, ...)
+void XbSymbolContext_SetContinuousSigScan(XbSymbolContextHandle pHandle, bool enable)
 {
-    char bufferTemp[2048];
+    iXbSymbolContext* pContext = (iXbSymbolContext*)pHandle;
 
-    // Check if output function is set.
-    // Plus if pass minimum verbose level to output.
-    if (output_func != NULL && mLevel >= output_verbose_level) {
-        va_list args;
-        va_start(args, format);
-        (void)vsprintf(bufferTemp, format, args);
-        va_end(args);
-
-        output_func(mLevel, bufferTemp);
+    if (iXbSymbolContext_AllowSetParameter(pContext)) {
+        pContext->one_time_scan = !enable;
     }
+
 }
 
-const char* XbSymbolLibraryToString(uint32_t library_flag)
+void XbSymbolContext_SetFirstDetectAddressOnly(XbSymbolContextHandle pHandle, bool enable)
+{
+    iXbSymbolContext* pContext = (iXbSymbolContext*)pHandle;
+
+    if (iXbSymbolContext_AllowSetParameter(pContext)) {
+        pContext->scan_first_detect = enable;
+    }
+
+}
+
+const char* XbSymbolDatabase_LibraryToString(uint32_t library_flag)
 {
     switch (library_flag) {
         case XbSymbolLib_D3D8: {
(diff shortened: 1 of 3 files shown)


RadWolfie:
Currently, multi-thread support is not enabled for MSVC due to the developers haven't implemented C11 support in their compiler. From what I had read in the recent past, they're working on it and should be bundle in future Visual Studio 2019 update.

RadWolfie:
Plus most importantly, include pointer to kernel thunk's address to translate internally for make stronger OOVPA signatures easier. 😄

RadWolfie:
If the team want to make the final decision for API's names later, please let me know here or on discord. Otherwise, we need to continue discussion what's need to be add/remove/update the parameters and/or API's names.

XbSymbolDatabase.h
```diff
@@ -447,6 +834,33 @@ static const char* xbe_type_str[XB_XBE_TYPE_MAX] = {
...
+    Library*        filters;
+} LibraryHeader;
+
+typedef struct _Section {
+    // These three variables must be fixed info.
+    char            name[8];
+    uint32_t        xb_virt_addr;
+    uint32_t        buffer_size;
+    void*           buffer_lower;
+} Section;
```
JayFoxRox: These very generic names should get prefixes to avoid conflicts with other APIs / user-code.

Ideally this was true (and consistent) for *all* types / variables / functions in the public headers.

XbSymbolDatabase.h
```diff
@@ -551,6 +951,73 @@ void XbSymbolContinuousSigScan(bool enable);
 /// <param name="enable">Input boolean to use first symbol address only or not.</param>
 void XbSymbolFirstDetectAddressOnly(bool enable);
 
+// ******************************************************************
+// * API v3 (context) functions to use with other projects.
```
JayFoxRox: Why even keep the old API? People who don't want to use the new API can just check out the old version (at loss of features), or make independent wrappers.
Porting apps is probably easy eitherway (and there aren't many users to begin with).

XbSymbolDatabase.h
```diff
@@ -551,6 +951,73 @@ void XbSymbolContinuousSigScan(bool enable);
...
+
+/// <summary>
+/// Step 3: Create XbSymbolContext context for the scan process.
+/// </summary>
+/// <param name="ppContext">Output XbSymbolContext handler.</param>
+/// <param name="register_func">Register callback function for any detected symbols during scan process.</param>
+/// <param name="library_input">See XbSymbolGenerate_LibraryFilter for details.</param>
+/// <param name="section_input">See XbSymbolGenerate_SectionFilter for details.</param>
+/// <param name="kernel_thunk">Input decoded virtual kernel thunk.</param>
+bool XbSymbol_Init(XbSymbolContext* ppContext, xb_symbol_register_t register_func, LibraryHeader library_input, SectionHeader section_input, xbaddr kernel_thunk);
```
JayFoxRox: Why do I `XbSymbol_Init` to *create*?
And why does `Init` happen as step 3?
This implies the name is bad.

XbSymbolDatabase.h
```diff
@@ -551,6 +951,73 @@ void XbSymbolContinuousSigScan(bool enable);
...
+
+/// <summary>
+/// Step 4: Perform a manual scan to set xreferences manually by requirement.
+/// </summary>
+/// <param name="pContext">Input XbSymbolContext handler.</param>
+void XbSymbolScan_Manual(XbSymbolContext pContext);
```
JayFoxRox: These new prefixes are weird?
Functions with different prefix interact; but unlike COM, there don't seem to be objects involved.

I'd expect this to be `XbSymbolContext_ScanManual` (`Object_Method`) or `XbSymbolScanManual` (`PrefixMethod` - existing style) or `XbSymbol_ScanManual` ('Module_Method')

XbSymbolDatabase.h
```diff
@@ -447,6 +834,33 @@ static const char* xbe_type_str[XB_XBE_TYPE_MAX] = {
...
+    uint32_t        buffer_size;
+    void*           buffer_lower;
+} Section;
+
+typedef struct _SectionHeader {
+    unsigned int    count;
+    Section*        filters;
+} SectionHeader;
+
+typedef void* XbSymbolContext;
```
JayFoxRox: This should be renamed to suggest that it is a handle or reference.

Current name implies a struct / class / object; so I'd expect code to work like this (which is incorrect):

```c
XbSymbolContext* ctx = malloc(sizeof(XbSymbolContext));
XbSymbol_Init(&XbSymbolContext, ...);
```

or (also incorrect):

```c
XbSymbolContext* ctx = XbSymbol_CreateContext(...);
```

In C, you usually assume a type to be an object that you don't want to assign / pass by value (as it might be too big). So if you have a non-struct-like-type, something should imply that it's a number / pointer of some sort.

XbSymbolDatabase.h
```diff
@@ -551,6 +951,73 @@ void XbSymbolContinuousSigScan(bool enable);
...
 void XbSymbolFirstDetectAddressOnly(bool enable);
 
+// ******************************************************************
+// * API v3 (context) functions to use with other projects.
+// ******************************************************************
+// NOTE: If need to make customize setup, please use the above functions
+// before call XbSymbol_Init function.
+
+/// <summary>
+/// Step 1: Generate library array for LibraryHeader input.
```
JayFoxRox: Ideally the API would imply the order of calls, so we don't have to document it.

XbSymbolDatabase.h
RadWolfie: True, I will fix it later on in this pull request.

XbSymbolDatabase.h
RadWolfie: There are necessary steps must be done first before pass on two data arrays. But yeah, I see your point for whole name is bad. It is still possible to manually create both data arrays on your own if using static library method.

XbSymbolDatabase.h
RadWolfie: Well, if use a document tool such as doxygen will generate a documentation for human readability as a help/research tool. Beside, if hover above the API name also provide details too.

RadWolfie: Well, if use a document tool such as doxygen will generate a documentation for human readability as a help/research tool. Beside, if hover above the API name also provide details too.

XbSymbolDatabase.h
RadWolfie: fixed

XbSymbolDatabase.h
RadWolfie: fixed

XbSymbolDatabase.h
```diff
@@ -551,6 +951,73 @@ void XbSymbolContinuousSigScan(bool enable);
...
+bool XbSymbolContext_Create(XbSymbolContextHandle* ppHandle, xb_symbol_register_t register_func, XbSDBLibraryHeader library_input, XbSDBSectionHeader section_input, xbaddr kernel_thunk);
+
+/// <summary>
+/// Step 4: Perform a manual scan to set xreferences manually by requirement.
+/// </summary>
+/// <param name="pHandle">Input XbSymbolContextHandle handler.</param>
+void XbSymbolContext_ScanManual(XbSymbolContextHandle pHandle);
+
+/// <summary>
+/// Step 5: (multi-thread) Process individual library input by third-party.
```
JayFoxRox: What makes this multi-thread?
Is it just thread-safe, or does it actually spawn threads internally?

And why is this part of the "Step" instead of generic information in the summary?

XbSymbolDatabase.h
```diff
@@ -551,6 +951,73 @@ void XbSymbolContinuousSigScan(bool enable);
...
+/// </summary>
+/// <param name="pHandle">Input XbSymbolContextHandle handler.</param>
+/// <param name="pLibrary">Input pointer of a library to start a scan process.</param>
+/// <param name="xref_first_pass">True: Optimized first search scan process; False: 2nd call and later will continue to return non-zero; see return for detail.</param>
+/// <returns>Return total xref count found. If return zero, then there's nothing left to find.</returns>
```
JayFoxRox: > If return zero, then there's nothing left to find.

This isn't grammatically correct.
Should this be or something like this:

> If it returns zero, then there's nothing left to find.

Could be made even better with more changes.

XbSymbolDatabase.h
```diff
@@ -551,6 +951,73 @@ void XbSymbolContinuousSigScan(bool enable);
...
+unsigned int XbSymbolContext_ScanLibrary(XbSymbolContextHandle pHandle, const XbSDBLibrary* pLibrary, bool xref_first_pass);
+
+/// <summary>
+/// Step 5: (single-thread) Process all of the libraries internally.
```
JayFoxRox: The single-thread note feels unnecessary.

XbSymbolDatabase.h
```diff
@@ -551,6 +951,73 @@ void XbSymbolContinuousSigScan(bool enable);
...
+/// </summary>
+/// <param name="pHandle">Input XbSymbolContextHandle handler.</param>
+void XbSymbolContext_ScanFull(XbSymbolContextHandle pHandle);
```
JayFoxRox: Maybe rename to `ScanAllLibraries` if that's what it does (or does it do anything else, other than scanning libs?).

XbSymbolDatabase.h
```diff
@@ -551,6 +951,73 @@ void XbSymbolContinuousSigScan(bool enable);
...
+
+/// <summary>
+/// Step 6: Register any xreferences may not had been output during the scan process.
```
JayFoxRox: Note: Sometimes you use xreferences (here) and sometimes xrefs.
For comments, I'd suggest calling them "cross-references" for clarity.

XbSymbolDatabase.h
JayFoxRox: I still think this implies bad design.

People can also learn the order by looking at the code sample, or it will be implied by what each function is doing (which implies the order).
For example, you can't possibly do: `XbSymbolContext_ScanManual` if you didn't create an object of type `XbSymbolContext` before (using `XbSymbolContext_Create` - which is a misnomer at the moment). This implies an order without having to explicitly document it.

In doxygen, you wouldn't be putting the step information near each function. Instead, you'd keep the generic documentation (order of calls) separated.

XbSymbolDatabase.h
```diff
@@ -551,6 +951,73 @@ void XbSymbolContinuousSigScan(bool enable);
...
+
+/// <summary>
+/// Step 3: Create XbSymbolContextHandle context for the scan process.
+/// </summary>
+/// <param name="ppHandle">Output XbSymbolContextHandle handler.</param>
+/// <param name="register_func">Register callback function for any detected symbols during scan process.</param>
+/// <param name="library_input">See XbSymbolDatabase_GenerateLibraryFilter for details.</param>
+/// <param name="section_input">See XbSymbolDatabase_GenerateSectionFilter for details.</param>
+/// <param name="kernel_thunk">Input decoded virtual kernel thunk.</param>
+bool XbSymbolContext_Create(XbSymbolContextHandle* ppHandle, xb_symbol_register_t register_func, XbSDBLibraryHeader library_input, XbSDBSectionHeader section_input, xbaddr kernel_thunk);
```
JayFoxRox: If you go with the `Object_Action` naming scheme, then this is a misnomer.
You'd want to call this `XbSymbolDatabase_CreateXbSymbolContext` or something (so it creates the object which can be used later to operate on).

XbSymbolDatabase.h
```diff
@@ -551,6 +951,73 @@ void XbSymbolContinuousSigScan(bool enable);
...
+unsigned int XbSymbolDatabase_GenerateLibraryFilter(const void* xb_header_addr, XbSDBLibraryHeader* library_out);
+
+/// <summary>
+/// Step 2: Generate section array for SectionHeader input.
+/// First call with <paramref name="section_out"/> as null pointer will return total count. Then second call will insert information to <paramref name="section_out"/>.filters field.
+/// </summary>
+/// <param name="xb_header_addr"></param>
+/// <param name="section_out">Input null pointer will provide the count to allocate memory, then allocate memory to section_out.filters field.</param>
+/// <param name="is_raw">True: Convert array to use raw xbe's relative address; False: Convert array to use xbox virtual memory relative address.</param>
+unsigned int XbSymbolDatabase_GenerateSectionFilter(const void* xb_header_addr, XbSDBSectionHeader* section_out, bool is_raw);
```
JayFoxRox: This stuff (filters) could probably be treated similar to the function which creates a context (returning a handle and share the naming conventions: `Create` or something).

Basically this would imply that you have to create the filters first, in order to be able to create the context (as it depends on the filters). Then you can operate on the context.

So the API will eventually become self-documenting.

XbSymbolDatabase.h
RadWolfie: Sorry, some of my changes got lost in the process. It's suppose to be `XbSymbolContextCreate`.

XbSymbolDatabase.h
RadWolfie: Multi-thread safe, if the compiler support C11 standard. Currently, msvc compiler doesn't have C11 standard support.

XbSymbolDatabase.h
RadWolfie: It's more of a scan all filter libraries. The proper wording for me would be `ScanAllFilterLibs`.

XbSymbolDatabase.h
RadWolfie: "x" wasn't used for "cross", it was actually "xbox" internal usage in the past codebase. And it still sort of is, I'll extend the comment about this.

XbSymbolDatabase.h
RadWolfie: Right now, there's no sample in this repo. Nor has a tool to for scan xbe files. That's the main reason steps are in the documentations. Later down the road, all of the steps comment will be remove.

XbSymbolDatabase.h
JayFoxRox: Even that name is somewhat unstandard and fragments the API further.

The proposed name would make this the only function that is *not* of style `Object_Action`.
It would also lead to the old issue with overlapping prefixes (`XbSymbol` and `XbSymbolContext`).

If we keep the old API, too (we shouldn't.. but this PR doesn't seem to remove it yet), then this is even more confusing, as that already used such naming conventions.

I'm not sure how a user could figure out `XbSymbolContextCreate` without looking at the headers (and even then it's confusing to me).

As a user, I'd have expected the call I have mentioned (if the rest of the API uses objects or a singleton).

XbSymbolDatabase.h
RadWolfie: I'm avoiding handle for both header's structures. Since I don't have a valid reason for one time usage which is only used in create process. And it's easier to reduce extra workload for simple allocation and set the new temporary buffer once. Once `XbSymbolContextCreate` have a copy of the filter's structure data. Then it is safe to free temporary buffer. Unless use the alternate function for multi-threading support, then you'll have to provide library info to `XbSymbolContext_ScanLibrary`.

Plus, I don't plan on implement a reference counter system.

XbSymbolDatabase.h
RadWolfie: 👌

XbSymbolDatabase.h
RadWolfie: It's use for awareness, beside I have reword it to understand better.

XbSymbolDatabase.h
RadWolfie: Updated in the comment section.

RadWolfie:
Since the API naming and comments was changed since initialize commit. I do know initialize commit is working and had been verified using my unit test project locally.

RadWolfie:
What's been change:

Added:
* XbSymbolDatabase_GetKernelThunkAddress

Renamed:
* XbSymbolContext_RegisterLibrary
* XbSymbolContext_SetBypassBuildVersionLimit
* XbSymbolContext_SetContinuousSigScan
* XbSymbolContext_SetFirstDetectAddressOnly
* XbSymbolDatabase_LibraryVersion
* XbSymbolDatabase_SetOutputMessage
* XbSymbolDatabase_LibraryToFlag
* XbSymbolDatabase_TestOOVPAs
* XbSymbolDatabase_SetOutputVerbosity

---

Internal functions has been renamed in second commit.

Plus added:
* iXbSymbolContext_Lock
* iXbSymbolContext_Unlock
* iXbSymbolContext_AllowSetParameter
* iXbSymbolContext_AllowScanLibrary

RadWolfie:
**NOTE:** `XbSymbolScan` function has not been remove yet. It is now deprecated function.

XbSymbolDatabase.c
```diff
@@ -84,59 +89,97 @@ static inline uint32_t BitScanReverse(uint32_t value)
...
+// doing the same scan process.
+typedef enum _eLibraryType {
+    LT_D3D = 0,
+    LT_AUDIO,
+    LT_JVS,
+    LT_XAPI,
+    LT_GRAPHIC,
+    LT_NETWORK,
+    LT_COUNT,
+    LT_UNKNOWN=LT_COUNT
```
PatrickvL: It's cleaner to swap these two, so list LT_UNKNOWN first, and declare LT_COUNT as an assignment of LT_UNKNOWN.

XbSymbolDatabase.c
```diff
@@ -359,21 +488,190 @@ uint32_t XbSymbolLibraryToFlag(const char* library_name)
...
+                        output_message_format(&output, XB_OUTPUT_MESSAGE_WARN,
+                            "D3D8 version 1.0.%d.%d Title Detected: This game uses an alias version 1.0.5788",
+                            xb_library_versions[library_index].wBuildVersion,
+                            xb_library_versions[library_index].wFlags.QFEVersion);
+
+                        library_header->filters[count].build_version = 5788;
+                    }
+                }
+            }
+            count++;
```
PatrickvL: This causes the `filters` array to have gaps. If that's undesirable, increase `count` only when populating a filter.

Same happens below, I'll mark them with a single `gap` comment.

XbSymbolDatabase.c
```diff
@@ -359,21 +488,190 @@ uint32_t XbSymbolLibraryToFlag(const char* library_name)
...
+        // Manual check if DSOUND section do exist.
+        if (!has_dsound_library) {
+            if (XbSymbolHasDSoundSection(xb_header_addr)) {
+                if (library_header != NULL) {
+                    library_header->filters[count].flag = XbSymbolLib_DSOUND;
+                    library_header->filters[count].build_version = xb_library_versions[library_index].wBuildVersion;
+                    library_header->filters[count].qfe_version = 0;
+                    memcpy(library_header->filters[count].name, Lib_DSOUND, 8);
+                }
+                count++;
```
PatrickvL: `gap`

XbSymbolDatabase.c
```diff
@@ -359,21 +488,190 @@ uint32_t XbSymbolLibraryToFlag(const char* library_name)
...
+
+        // Manual check if Xbe type is debug or Chihiro
+        if (XbeType != XB_XBE_TYPE_RETAIL) {
+            if (library_header != NULL) {
+                library_header->filters[count].flag = XbSymbolLib_JVS;
+                library_header->filters[count].build_version = xb_library_versions[library_index].wBuildVersion;
+                library_header->filters[count].qfe_version = 0;
+                memcpy(library_header->filters[count].name, Lib_JVS, 8);
+            }
+            count++;
```
PatrickvL: `gap`

XbSymbolDatabase.c
```diff
@@ -359,21 +488,190 @@ uint32_t XbSymbolLibraryToFlag(const char* library_name)
...
+
+                        if (is_raw) {
+                            sv_index->buffer_lower = (memptr_t)xb_header_addr + sh_index->dwRawAddr;
+                        }
+                        else {
+                            sv_index->buffer_lower = xb_start_addr + sh_index->dwVirtualAddr;
+                        }
+                    }
+
+                    count++;
```
PatrickvL: `gap`

XbSymbolDatabase.c
```diff
@@ -84,59 +89,97 @@ static inline uint32_t BitScanReverse(uint32_t value)
...
-
-    OOVPATable     *OovpaTable;
+    OOVPATable*     OovpaTable;
     unsigned int    OovpaTableCount;
 } SymbolDatabaseList;
 
 SymbolDatabaseList SymbolDBList[] = {
     // Support inline functions in .text section
-    { XbSymbolLib_D3D8 | XbSymbolLib_D3D8LTCG,{ Sec_text, Sec_D3D, Sec_FLASHROM }, D3D8_OOVPA, D3D8_OOVPA_COUNT },
+    { LT_D3D, XbSymbolLib_D3D8 | XbSymbolLib_D3D8LTCG,{ Sec_text, Sec_D3D, Sec_FLASHROM }, D3D8_OOVPA, D3D8_OOVPA_COUNT },
```
PatrickvL: Why include a library type enum value here, when there's also a `internal_GetLibraryType` function that's doing the same? Either remove this variable, and use the function, or remove the function (or implement it as a lookup in SymbolDBList, instead of switch/case code)

XbSymbolDatabase.c
```diff
@@ -658,39 +956,113 @@ void XbSymbolScanOOVPA(OOVPATable *OovpaTable,
...
+            return LT_XAPI;
+        case XbSymbolLib_XGRAPHC:
+            return LT_GRAPHIC;
+        case XbSymbolLib_XONLINE:
+        case XbSymbolLib_XONLINES:
+        case XbSymbolLib_XNET:
+        case XbSymbolLib_XNETS:
+        case XbSymbolLib_XNETN:
+            return LT_NETWORK;
+    }
```
PatrickvL: Remove this function, or implement it as a lookup in `SymbolDBList` (instead of switch/case code).

XbSymbolDatabase.c
```diff
@@ -762,158 +1134,253 @@ bool XbSymbolScanSection(uint32_t xbe_base_address,
...
+    pContext->section_input.count = section_input.count;
+    pContext->section_input.filters = malloc(sizeof(XbSDBSection) * section_input.count);
+    if (pContext->section_input.filters == NULL) {
+        goto LibraryCleanup;
+    }
+
+    memcpy(pContext->library_input.filters, library_input.filters, sizeof(XbSDBLibrary) * library_input.count);
+    memcpy(pContext->section_input.filters, section_input.filters, sizeof(XbSDBSection) * section_input.count);
+
+                            // Mark all Xrefs initially as undetermined
```
PatrickvL: nit : indentation

XbSymbolDatabase.h
```diff
@@ -91,12 +91,393 @@ extern "C" {
...
+    KT_IMP_FUNC_XProfpControl,                          // 0x0172 (370) PROFILING
+    KT_IMP_FUNC_XProfpGetData,                          // 0x0173 (371) PROFILING
+    KT_IMP_FUNC_IrtClientInitFast,                      // 0x0174 (372) PROFILING
+    KT_IMP_FUNC_IrtSweep,                               // 0x0175 (373) PROFILING
+    KT_IMP_FUNC_MmDbgAllocateMemory,                    // 0x0176 (374) DEVKIT ONLY!
+    KT_IMP_FUNC_MmDbgFreeMemory,                        // 0x0177 (375) DEVKIT ONLY!
+    KT_IMP_FUNC_MmDbgQueryAvailablePages,               // 0x0178 (376) DEVKIT ONLY!
+    KT_IMP_FUNC_MmDbgReleaseAddress,                    // 0x0179 (377) DEVKIT ONLY!
+    KT_IMP_FUNC_MmDbgWriteCheck,                        // 0x017A (378) DEVKIT ONLY!
+    KT_IMP_COUNT,
```
PatrickvL: Instead of `KT_IMP_` prefix, use `XREF_KT_` prefix.

XbSymbolDatabase.c
```diff
@@ -1430,178 +1869,252 @@ bool XbLibraryScan(custom_scan_func_t custom_scan_func,
...
+    XbSymbolContextHandle pHandle;
+    iXbSymbolContext* iContext;
+    XbSDBLibraryHeader library_input;
+    XbSDBSectionHeader section_input;
+
+    // Step 1, let's get the total sum of array to allocate library input.
+    library_input.count = XbSymbolDatabase_GenerateLibraryFilter(xb_header_addr, NULL);
+    // If total sum is zero, then the input is invalid.
+    if (library_input.count == 0) {
+        goto EmptyCleanup;
```
PatrickvL: Avoid goto, just return here.

XbSymbolDatabase.c
```diff
@@ -1430,178 +1869,252 @@ bool XbLibraryScan(custom_scan_func_t custom_scan_func,
...
+    }
+    // Then allocate the necessary memory requirement to obtain the information.
+    library_input.filters = malloc(sizeof(XbSDBLibrary) * library_input.count);
+    // Verify if memory has been allocated.
+    if (library_input.filters == NULL) {
+        goto EmptyCleanup;
```
PatrickvL: Avoid goto, just return here.

XbSymbolDatabase.c
```diff
@@ -658,39 +956,113 @@ void XbSymbolScanOOVPA(OOVPATable *OovpaTable,
...
+        case XbSymbolLib_XNETN:
+            return LT_NETWORK;
+    }
+}
+
+xbaddr XbSymbolDatabase_GetKernelThunkAddress(const void* xb_header_addr)
+{
+    xb_xbe_type xbe_type = GetXbeType(xb_header_addr);
+
+    uint32_t kernel_thunk_addr = *(uint32_t*)((uint8_t*)xb_header_addr + 0x0158);
```
PatrickvL: Use a symbol instead of the 'magic' literal hexadecimal number `0x0158`.

XbSymbolDatabase.c
RadWolfie: Okay, I'll need to take a step back to see what you're seeing. I'm sure your suggestion method is better. 😁

XbSymbolDatabase.c
RadWolfie: I think I see what you mean, I'll re-review this function again.

XbSymbolDatabase.c
RadWolfie: I didn't make any modification here except move into its own function.

XbSymbolDatabase.c
RadWolfie: After reviewing again, obviously this variable is no longer needed. 😃

XbSymbolDatabase.c
RadWolfie: Since SymbolDBList could be used two or more times in a single library scan. This function, `internal_GetLibraryType` will stay to keep the scan process active once. 😉

XbSymbolDatabase.c
RadWolfie: I don't see the problem. I think you had missed the API's documentation. If `library_header` is null pointer, it will return the counts for what is needed to allocate filter's array. Then re-call the function again to fill in the filters.

RadWolfie:
Fixed all reviews by @PatrickvL except one which is working as intended.

Plus I discover a warning message from Visual Studio about memcpy usage over the limit. I have replaced memcpy to strncpy function for prevent warning and null'd the overflow.

PR #95, opened 2019-10-29
Derive XREF_OFFSET_D3DDEVICE_M_VERTEXSHADER

PatrickvL:
Derive XREF_OFFSET_D3DDEVICE_M_VERTEXSHADER (needs offset verification), and register it's address as "D3DDevice__m_VertexShader" (required for VertexShader unpatching in Cxbx-Reloaded)


OOVPADatabase/D3D8/3911.inl
```diff
@@ -24,44 +24,45 @@
 // ******************************************************************
 
 // ******************************************************************
-// * D3DDevice__ManualFindEventHandleGeneric
+// * D3DDevice__m_VerticalBlankEvent__ManualFindGeneric
```
JayFoxRox: Feels like these changes should be split into a separate commit or PR.
It's not described in the PR description and hides the other changes.

RadWolfie:
Respectfully, I don't plan on fix rebase on top of this change for API v3.0's #94 pull request. My suggestion is to split into 2 pull requests on top of API v3.0, #94 pull request.

PatrickvL:
@RadWolfie Ofcourse it's possible to base those PR on something else, like the branch you mention.

You'll have to forgive me not noticing what you point at. It's 3 months old work - what's the current state?

RadWolfie:
> You'll have to forgive me not noticing what you point at. It's 3 months old work - what's the current state?

Has been functional, just need to finalize the API's names and/or update the parameters for worthy options to have.

RadWolfie:
Base on offset's location from 3911's signature and later. Standard D3D8's m_VertexShader are correct since they are at same offset location as in the signatures' database.

However, most of LTCG's m_VertexShader offsets are incorrect. Here's what I have discover by comparing title with 3925, standard D3D8's signatures, and D3D8LTCG's signatures.

D3DDevice_SetVertexShader_0 (LTCG)
* 3911
  * **0x0D** = XREF_OFFSET_D3DDEVICE_M_VERTEXSHADER **(confirmed)**
* 4039
  * **0x12** = XREF_OFFSET_D3DDEVICE_M_VERTEXSHADER **(unverified, yet should be align base on existing 4 bytes in signatures)**
* 4432
  * **0x13** = XREF_OFFSET_D3DDEVICE_M_VERTEXSHADER **(OK, yet unverified)**
* 5233
  * **0x0E** = XREF_OFFSET_D3DDEVICE_M_VERTEXSHADER **(unverified, yet should be align base on existing 4 bytes in signatures)**

The "existing 4 bytes in signatures" are **0x55, 0x56, 0x8B, and 0x35** (in order). These 4 bytes are align to where **XREF_PD3DDEVICE** is located. After this is where the member of vertex shader offset assembly command can be found.

PatrickvL:
@RadWolfie Did you verify the 5455 (non-LTCG) offset too? (It's marked with a TODO, see https://github.com/Cxbx-Reloaded/XbSymbolDatabase/blob/d605698eddd3086e627b7053d99e193d47538b36/OOVPADatabase/D3D8/5455.inl#L110 )

RadWolfie:
Yes, even 5455 (non-LTCG) is correct too.

OOVPADatabase/D3D8/3911.inl
JayFoxRox: Most of this seems to be in PR 96 now.
This PR still needs a temporary rebase so the commit loses that information.

PatrickvL: Yeah, I'm busy with that

PatrickvL: Done. I'll move the rest of the changes in separate commits / PR's as well.

PR #99, opened 2019-12-10
HOTFIX: Fix Unix Builds

RadWolfie:
Recently discovered not all compilers support C11's thread support. Although, only needed mutex support.


RadWolfie:
Low risk merge

PR #102, opened 2019-12-14
Fix DSound VTable Not Found

### LINKED ISSUE #88: DSOUND Stream VTable symbols missing
I have discover a bug while verifying Xbox Live xbe's symbols are detecting correctly or not.

Base on `XREF_DSS_VOICE_VTABLE` and `XREF_DSS_STREAM_VTABLE` virtual addresses, they are reside in `.text` section instead of `.rdata` section (which doesn't exist). Which means certain symbols will not be registered such as:
- CDirectSoundStream_AddRef
- CDirectSoundStream_Release
- CDirectSoundStream_GetInfo
- ... and so on.

Further investigation is needed to avoid any conflict with certain OOVPAs which is/are, and always is, weak to scan in `.text` section.

---

**General titles shouldn't be affected by this.**

RadWolfie:
Close #88 

Xbox Live now does register missing references using Cxbx-Reloaded's custom develop build link to this pull request.

2nd commit is a discovery which were preventing DSound's 2nd (actual) OOVPA database to scan.

### DIFF
--- a/OOVPADatabase/DSound_OOVPA.inl
+++ b/OOVPADatabase/DSound_OOVPA.inl
@@ -161,6 +161,13 @@
 #include "DSound/5455.inl"
 #include "DSound/5558.inl"
 
+// ******************************************************************
+// * DSOUND_OOVPA_manual
+// ******************************************************************
+OOVPATable DSound_OOVPA_manual[] = {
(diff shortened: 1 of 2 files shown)


RadWolfie:
Merging as-is

PR #104, opened 2020-02-08
Update pr_verify_build.yml

Margen67:
Add push trigger.
Deduplicate steps with build matrix.
Ignore commits to README.md.
Use v2 instead of master for checkout action.


RadWolfie:
Nice work! Although, I don't see a reason to have a push trigger for any branches in this repo. The main reason of having CI to test against pull requests only, is well... We expect pull requests only and the main branch is protected.

PR #106, opened 2020-09-03
Build fixes

thrimbor:
These commits fix some build issues Matt and me encountered when building XbSymbolDatabase on Linux (Arch in my case, and Ubuntu I assume for Matt's). These are just to make the code build correctly, I plan to upstream Matt's cli tool and addition XONLINE signatures soon.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,11 +1,9 @@
 # CMakeList.txt : CMake project for XbSymbolSymbolDatabase, include source and define
 # project specific logic here.
 #
-cmake_minimum_required (VERSION 3.12)
+cmake_minimum_required (VERSION 3.10.2)
 
 project (XbSymbolDatabase)
-
-add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
 
 file(GLOB_RECURSE SOURCES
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # CMakeList.txt : CMake project for XbSymbolSymbolDatabase, include source and define
 # project specific logic here.
 #
-cmake_minimum_required (VERSION 3.12)
+cmake_minimum_required (VERSION 3.10.2)
```
RadWolfie, line 4: Unfortunately, 3.12 is the minimal requirement due to `add_compile_definitions` function. Since it doesn't exist and certain CMake variant will make a fatal error by enforcing check what is used in here.

If there's a necessary to lower the minimal requirement, then an alternative solution is require before accept into upstream.

RadWolfie (changes requested):

CMakeLists.txt
thrimbor: Replaced `add_compile_definitions` with `target_compile_definitions` so that we're 3.10.2 compatible now.

RadWolfie (approved):

PR #108, opened 2020-09-03
Add a CLI tool and build it with GH actions

thrimbor:
This adds a cli tool, which can be used to run symbol detection on an XBE. The actual tool was written by @mborgerson, I only tweaked the CMakeLists to fix building under some Linux distributions and to make building the tool optional (as Cxbx-Reloaded probably doesn't want it built). The Actions file is also his work, I slightly modified it to ignore the readme (like the current Actions file does) and make it use checkout@v2.

The Actions file also uploads the created CLI tool binaries as an artifact. This is then used in @mborgerson's Ghidra XBE plugin, where this artifact is downloaded and included to provide symbol detection.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,16 +4,26 @@ cmake_minimum_required (VERSION 3.10.2)
 
 project (XbSymbolDatabase)
 
-if(NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
- option(XBSDB_INSTALL_LIB "Install libXbSymbolDatabase library" OFF)
+if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+ set(XBSDB_DEFAULT_CONFIGS ON)
 else()
- set(XBSDB_INSTALL_LIB ON)
+ set(XBSDB_DEFAULT_CONFIGS OFF)
 endif()
+
+option(XBSDB_INSTALL_LIB "Install libXbSymbolDatabase library" ${XBSDB_DEFAULT_CONFIGS})
+option(XBSDB_BUILD_CLI "Build XbSymbolDatabase CLI tool" ${XBSDB_DEFAULT_CONFIGS})
+include(CMakeDependentOption)
+cmake_dependent_option(XBSDB_INSTALL_CLI "Install XbSymbolDatabase CLI tool" ${XBSDB_DEFAULT_CONFIGS}
+                       "XBSDB_BUILD_CLI" OFF)
 
 # For any optional tools, use list(APPEND XBSDB_INSTALL_TOOLS "tool_name")
 
 if(XBSDB_INSTALL_LIB)
  list(APPEND XBSDB_INSTALL_TOOLS "libXbSymbolDatabase")
+endif()
+
+if(XBSDB_INSTALL_CLI)
+ list(APPEND XBSDB_INSTALL_TOOLS "XbSymbolDatabaseCLI")
 endif()
 
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
--- a/projects/CMakeLists.txt
+++ b/projects/CMakeLists.txt
@@ -3,3 +3,7 @@
 cmake_minimum_required (VERSION 3.10.2)
 
 add_subdirectory("libXbSymbolDatabase")
+
+if(XBSDB_BUILD_CLI)
+ add_subdirectory("cli")
+endif()
--- a/projects/cli/CMakeLists.txt
+++ b/projects/cli/CMakeLists.txt
@@ -1,0 +1,19 @@
+# CMakeList.txt : CMake project for XbSymbolDatabaseCLI.
+#
+cmake_minimum_required(VERSION 3.10.2)
+
+project(XbSymbolDatabaseCLI LANGUAGES C)
+
+find_package(Threads)
+
+add_executable(XbSymbolDatabaseCLI ${XBSDB_ROOT_DIR}/src/cli/main.c)
+
+target_compile_definitions(XbSymbolDatabaseCLI PRIVATE _CRT_SECURE_NO_WARNINGS)
+
+target_link_libraries(XbSymbolDatabaseCLI libXbSymbolDatabase Threads::Threads)
+
+set_target_properties(XbSymbolDatabaseCLI PROPERTIES
+ C_STANDARD 11
+ C_STANDARD_REQUIRED ON
+ FOLDER XbSymbolDatabase
+)
--- a/src/cli/main.c
+++ b/src/cli/main.c
@@ -1,0 +1,103 @@
+#include <stdio.h>
+#include <assert.h>
+#include <string.h>
+#include <libXbSymbolDatabase.h>
+
+static void msg_cb(xb_output_message message_flag, const char *msg)
+{
+    fprintf(stderr, "XbSymbolDatabase: %s\n", msg);
+}
+
+static void reg_cb(const char* library_str, uint32_t library_flag, const char* symbol_str, xbaddr address, uint32_t build_verison)
+{
+    printf("%.8s__%s = 0x%08x\n", library_str, symbol_str, address);
+}
+
+int main(int argc, char const *argv[])
+{
+    int verbose = 0;
+    const char *xbe_path = NULL;
+    int arg_off = 1;
+
+    if (argc > 1 && strcmp(argv[1], "-verbose") == 0) {
+        verbose = 1;
+        arg_off++;
+    }
+
+    if (arg_off >= argc) {
+        fprintf(stderr, "usage: %s [-verbose] xbefile\n", argv[0]);
+        return 1;
+    } else {
+        xbe_path = argv[arg_off];
+    }
+
+    unsigned int version = XbSymbolDatabase_LibraryVersion();
+    if (verbose) {
+        fprintf(stderr, "Database Version = %x\n", version);
+    }
+
+    FILE *fd = fopen(xbe_path, "rb");
+    if (fd == NULL) {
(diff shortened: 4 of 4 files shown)


RadWolfie:
I do not plan to add artifacts, which is temporary storage. It will only hold for 3 months. build.yml file doesn't make sense to me nor is optimized. Please restore the workflow then simply just add `-D BUILD_TOOL=ON` within cmake-parameters variable.

They can build their own binaries then make a release in order to store permanently. Currently, there are a lot of changes that needs to be done plus optimization.

P.S. I don't have time to at least possibility setup a semantic versioning to handle all of this work in order to make progressive releases.

CMakeLists.txt
```diff
@@ -15,6 +15,14 @@ source_group(TREE ${CMAKE_CURRENT_LIST_DIR} FILES ${SOURCES})
 add_library (XbSymbolDatabase ${SOURCES})
 target_compile_definitions(XbSymbolDatabase PRIVATE _CRT_SECURE_NO_WARNINGS)
 
+option(BUILD_TOOL "Build XbSymbolDatabase CLI tool" OFF)
+if(BUILD_TOOL)
```
RadWolfie, line 19: Keep in mind, I will move this changes into projects folder.

The goal for each project, such as library, cli tool, and unit test cmake files go into **projects** folder. Plus source files will likely go into their own folder due to multiple projects bundle in one repository. i.e.
```
src/cli/...
src/lib/...
src/unittest/...
```
or without src part.

thrimbor:
> I do not plan to add artifacts, which is temporary storage. It will only hold for 3 months.

I'm aware it's temporary, but I don't see how that's a problem. Rolling releases would probably be better, but artifacts at least are enough for building the Ghidra XBE plugin, and also allow people to, say, test the changes done in a PR locally without having to build themselves.

> build.yml file doesn't make sense to me

What about it doesn't make sense?

> They can build their own binaries then make a release in order to store permanently.

Are you saying that other projects should maintain their own forks of XbSymbolDatabase and add this functionality there? If yes, why? Imho this should be solved upstream instead.

RadWolfie:
> I'm aware it's temporary, but I don't see how that's a problem. Rolling releases would probably be better, but artifacts at least are enough for building the Ghidra XBE plugin, and also allow people to, say, test the changes done in a PR locally without having to build themselves.

I don't see the problem for fork repos to have their own artifacts and it's not a big issue either. We do build ourselves to verify the result. Since not every compiler/platform will build the same. And I have to pull the pull request branch locally to able verify with local, major wip, unit test tool.

> What about it doesn't make sense?

1. It's not optimized.
2. Has hardcode hacks which I don't accept.
3. Doesn't make sense to rename them to tool_name.____64.____. Since I don't see the purpose for this. Putting them in a folder name would be more common sense to me. (Same as reason 2.)
4. Too much effort to maintenance per platform.
5. Easy to miss corrections for all three or more platforms. (Same as reason 4.)

> Are you saying that other projects should maintain their own forks of XbSymbolDatabase and add this functionality there? If yes, why? Imho this should be solved upstream instead.

That's not what I'm saying, they can create a new branch, with workflow, then simply make merge commit that contain latest changes to produce pushed artifacts "temporary".

.github/workflows/build.yml
```diff
@@ -0,0 +1,77 @@
...
+      - uses: actions/checkout@v2
+      - name: Generate cmake files
+        shell: cmd
+        run: |
+          mkdir build && cd build
+          cmake .. -G "Visual Studio 16 2019" -A Win32 -D BUILD_TOOL=ON
+      - name: Build
+        run: |
+          cmake --build build/ --config ${{ matrix.configuration }}
+          Rename-Item -Path "build\${{ matrix.configuration }}\XbSymbolDatabaseTool.exe" -NewName "XbSymbolDatabaseTool.win64.${{ matrix.configuration }}.exe"
```
RadWolfie, line 28: `-A Win32` will output 32-bit executable on build, not 64-bit executable by the way.

cli.c
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    XbSDBLibraryHeader lib_hdr;
+    lib_hdr.count = XbSymbolDatabase_GenerateLibraryFilter(xbe, NULL);
+    lib_hdr.filters = malloc(lib_hdr.count * sizeof(XbSDBLibrary));
+    assert(lib_hdr.filters != NULL);
+    XbSymbolDatabase_GenerateLibraryFilter(xbe, &lib_hdr);
+
+    if (verbose) {
+        fprintf(stderr, "%d libraries\n", lib_hdr.count);
```
RadWolfie, line 72:
```suggestion
        fprintf(stderr, "%u libraries\n", lib_hdr.count);
```
Proper format for unsigned integer is %u.

cli.c
```diff
@@ -0,0 +1,103 @@
...
+    XbSDBSectionHeader sec_hdr;
+    sec_hdr.count = XbSymbolDatabase_GenerateSectionFilter(xbe, NULL, 1);
+    sec_hdr.filters = malloc(sec_hdr.count * sizeof(XbSDBSection));
+    assert(sec_hdr.filters != NULL);
+    XbSymbolDatabase_GenerateSectionFilter(xbe, &sec_hdr, 1);
+
+    if (verbose) {
+        fprintf(stderr, "%d sections\n", sec_hdr.count);
+        for (int i = 0; i < sec_hdr.count; i++) {
+            fprintf(stderr, "  %s\n", sec_hdr.filters[i].name);
```
RadWolfie, line 87: Same ideal with lib_hdr's filter[x].name need fixed size length for safe output.

RadWolfie (changes requested):
I found some changes require in cli's source code. The rest looks good.

thrimbor:
> > I'm aware it's temporary, but I don't see how that's a problem. Rolling releases would probably be better, but artifacts at least are enough for building the Ghidra XBE plugin, and also allow people to, say, test the changes done in a PR locally without having to build themselves.
> 
> I don't see the problem for fork repos to have their own artifacts and it's not a big issue either. We do build ourselves to verify the result. Since not every compiler/platform will build the same. And I have to pull the pull request branch locally to able verify with local, major wip, unit test tool.

I do - it requires forking a repo to add functionality that should be provided upstream. And ofc maintainers have to check the code out locally, but there's no need to force everybody else to do the same.

>     2. Has hardcode hacks which I don't accept.

Can you please be more specific?

> > Are you saying that other projects should maintain their own forks of XbSymbolDatabase and add this functionality there? If yes, why? Imho this should be solved upstream instead.
> 
> That's not what I'm saying, they can create a new branch, with workflow, then simply make merge commit that contain latest changes to produce pushed artifacts "temporary".

I'm confused - you say it's not what you were saying, but then you describe what I just said.

> In my opinion, learning more about the matrix method would be ideal how

I'm just going to ignore that condescending statement.

RadWolfie:
> I do - it requires forking a repo to add functionality that should be provided upstream. And ofc maintainers have to check the code out locally, but there's no need to force everybody else to do the same.

Who is everybody else? There are little maintenance happen here. Last updated before you create pull requests were 7 months ago.

> Can you please be more specific?

I'm not going to continue with this discussion. If you want my help to fix the workflow optimization, then let's talk in discord.

> I'm confused - you say it's not what you were saying, but then you describe what I just said.

The way you described is forked repository users will continue to do their work and not upstream plus "add this functionality". That's how I read it from your reply.
 
> I'm just going to ignore that condescending statement.

It's a suggestion by the way. It's too much work for us to maintenance without using matrix method. Beside it should be in separate pull request anyway.

.github/workflows/build.yml
```diff
@@ -0,0 +1,77 @@
...
+  windows:
+    name: Windows
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        configuration: [Debug, Release]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Generate cmake files
+        shell: cmd
```
Margen67, line 21: This line isn't needed.

.github/workflows/build.yml
```diff
@@ -0,0 +1,77 @@
...
+jobs:
+  windows:
+    name: Windows
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        configuration: [Debug, Release]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Generate cmake files
```
Margen67, line 20:
```suggestion
      - name: Generate CMake files
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,77 @@
...
+        shell: cmd
+        run: |
+          mkdir build && cd build
+          cmake .. -G "Visual Studio 16 2019" -A Win32 -D BUILD_TOOL=ON
+      - name: Build
+        run: |
+          cmake --build build/ --config ${{ matrix.configuration }}
```
Margen67, line 27:
```suggestion
          working-directory: build
          cmake --build . --config ${{ matrix.configuration }} -j $env:NUMBER_OF_PROCESSORS
```

Margen67, line 27:
```suggestion
        working-directory: build
        run: |
          cmake --build . --config ${{ matrix.configuration }} -j $env:NUMBER_OF_PROCESSORS
```

RadWolfie:
Since I have permission to fix this pull request. However, local push to this pull request isn't working for me. You'll need to do hard reset for 2-3 commits. Then pull rebased branch, from https://github.com/RadWolfie/XbSymbolDatabase/tree/clitool-fixup, with fixes from my fork repo.

I have pinged this message to @thrimbor on discord. Yet I had not seen any activity from this pull request, so I'm posting it here.

thrimbor:
I integrated your changes now.

RadWolfie (approved):
All checks are passing, comparison with containing fixes do match, and verified CLIs are built and installed base on artifact folder.

PR #111, opened 2020-09-05
Convert PR Workflow to CI Build And add Artifacts Process

RadWolfie:
Since @mborgerson requested temporary artifacts storage, which will only hold up to 3 months. Semantic Versioning releases, for permanent storage, will be implement later on.

Changes were made to use CI workflow for include push event instead of pull request event only. Plus use cmake's install method without need for per platform commands to copy files after build. Along with switch from Windows 32-bit executable to 64-bit executable.


RadWolfie:
@Margen67 When I add `-j $env:NUMBER_OF_PROCESSORS` for windows' matrix. It kept failing because it's not defined. Do you want to take look into this? Or don't try fix it?

Margen67:
> 
> 
> @Margen67 When I add `-j $env:NUMBER_OF_PROCESSORS` for windows' matrix run command. It kept failing because it's not defined. Do you want to take look into this? Or don't try fix it?

It should work for `--build`, otherwise I have no idea.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,56 @@
...
+        run: |
+          mkdir build && cd build
+          cmake .. ${{ matrix.cmake-parameters }}
+      - name: Build
+        working-directory: build
+        run: cmake --build . --config ${{ matrix.configuration }}
+      - name: Copy Files
+        if: matrix.configuration == 'Release'
+        working-directory: build
+        run: cmake --install . --config ${{ matrix.configuration }} --prefix ../upload/${{ matrix.folder }}
```
Margen67, line 39: If you change it to this you won't need the folder include:
```suggestion
        run: cmake --install . --config ${{ matrix.configuration }} --prefix ../upload/${{ runner.os }}64
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,56 @@
...
+        configuration: [Debug, Release]
+        include:
+          - os: windows-latest
+            cmake-parameters: -G "Visual Studio 16 2019" -A x64
+            folder: win64
+          - os: ubuntu-latest
+            cmake-parameters: -G "Unix Makefiles"
+            folder: linux64
+          - os: macos-latest
+            cmake-parameters: -G "Xcode"
```
Margen67, line 25: You only need quotes if it has spaces, but they can be left there for consistency.
```suggestion
            cmake-parameters: -G Xcode
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,56 @@
+name: GitHub CI
+
+on:
+  push:
+  pull_request:
+    paths-ignore:
+      - 'README.md'
```
Margen67, line 7:
```suggestion
      - '*.md'
      - '.github/*'
      - '.github/ISSUE_TEMPLATE/**'
```

.github/workflows/ci.yml
RadWolfie: What about other architectures? I would like to have both platforms and architecture combination remain possible than hardcode specific architecture. Which means if I do your way will have to revert back and make sure nothing else is broken.

I may will add ARM architecture in the future if there are requests for it.

[Also runner.os from the doc says it would be Windows, Linux, macOS](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#runner-context) So, there's no shorter name at runtime.

.github/workflows/ci.yml
RadWolfie: I think it's better to add `'.github/ISSUE_TEMPLATE/**'` to push's paths-ignore as well.

RadWolfie: > It should work for `--build`, otherwise I have no idea.

Weird, I retried it again. Now it's working. Except it doesn't work with other platforms.

RadWolfie:
> It should work for `--build`, otherwise I have no idea.

Weird, I retried it again. Now it's working. Except it doesn't work with other platforms.

.github/workflows/ci.yml
```diff
@@ -18,23 +18,26 @@ jobs:
...
       - uses: actions/checkout@v2
       - name: Generate CMake Files
         run: |
           mkdir build && cd build
-          cmake .. ${{ matrix.cmake-parameters }}
+          cmake .. ${{ matrix.cmake-generator }}
       - name: Build
         working-directory: build
-        run: cmake --build . --config ${{ matrix.configuration }}
+        run: cmake --build . --config ${{ matrix.configuration }} ${{ matrix.cmake-build-param }}
```
RadWolfie, line 42: With cmake-build-param option, msvc can build faster without need to make separate step.

Margen67:
> 
> 
> > It should work for `--build`, otherwise I have no idea.
> 
> Weird, I retried it again. Now it's working, except it doesn't work with other platforms.

I forgot that Linux/macOS require a different way of getting the number of CPU cores.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
...
+        os: [Windows-latest, Ubuntu-latest, macOS-latest]
+        configuration: [Debug, Release]
+        include:
+          - os: windows-latest
+            cmake-generator: -G "Visual Studio 16 2019" -A x64
+            cmake-build-param: -j $env:NUMBER_OF_PROCESSORS
+            folder: win64
+          - os: ubuntu-latest
+            cmake-generator: -G "Unix Makefiles"
+            cmake-build-param: ${{ null }}
```
Margen67, line 28:
```suggestion
            cmake-build-param: -j $(nproc --all)
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
...
+            folder: linux64
+          - os: macos-latest
+            cmake-generator: -G "Xcode"
+            cmake-build-param: ${{ null }}
```
Margen67, line 32:
```suggestion
            cmake-build-param: -j $(sysctl -n hw.ncpu)
```

.github/workflows/ci.yml
Margen67: You could add `arch`, but that might be too convoluted.

RadWolfie:
For comparison difference between Ubuntu/macOS builds of max processors set. Keep in mind, builds over time does not give same result on GitHub Actions.
Ubuntu:
* [before (6 seconds)](https://github.com/Cxbx-Reloaded/XbSymbolDatabase/runs/1076899571)
* [after (9 seconds)](https://github.com/Cxbx-Reloaded/XbSymbolDatabase/runs/1078324649)

macOS:
* [before (26 seconds)](https://github.com/Cxbx-Reloaded/XbSymbolDatabase/runs/1076899582)
* [after (49 seconds)](https://github.com/Cxbx-Reloaded/XbSymbolDatabase/runs/1078324661)

.github/workflows/ci.yml
RadWolfie: Hmm... different architectures will need go into a different job section in order for it to work which require duplicate steps. Or by using different os name which I'm not really interest in.

RadWolfie: Resolved

Instead of `64`, it will be `_x64` prefix. To separate platform and architecture readability.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
+name: GitHub CI
+
+on:
+  push:
+    paths-ignore:
+      - '.github/*TEMPLATE/**'
```
Margen67, line 6:
```suggestion
      - '*.md'
      - '.github/*'
      - '.github/*TEMPLATE/**'
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+
+jobs:
+  build-x64:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-latest, ubuntu-latest, macos-latest]
+        configuration: [Debug, Release]
+        include:
```
Margen67, line 21: Why not add the x86 job here instead of making it a separate one?

.github/workflows/ci.yml
RadWolfie: Any markdown changes should be pushed into artifacts, in case of important documentation.

.github/workflows/ci.yml
RadWolfie: I described the issues in previous discussion: https://github.com/Cxbx-Reloaded/XbSymbolDatabase/pull/111#discussion_r484204605

Include section does not allow any array nor will read duplicate os string name check. It causes to skip whatever is ordered first and only read the last one for same os name. In order to have same os string is to put it into a separate job section.

Margen67: What if you made the include for `cmake-generator` instead of `os`? (The ugly name could be fixed by using `name`)

RadWolfie: Is these lines in the link is what you're suggesting?
https://github.com/RadWolfie/XbSymbolDatabaseTestStuff/blob/4fe470701c2f1cc8bddc5d085364a5aa9cfbcf08/.github/workflows/ci.yml#L14-L54

Margen67: Yes

RadWolfie: Okay, I changed the workflow to the new solution.

Margen67: `plat` can be replaced with `runner.os`

RadWolfie: Can we rename runner.os to another name such as Android or Xbox One? 😉 I don't think GitHub Actions will have such feature. I do know CMake do have ability compile for other platforms. How? I'm not sure since their generator documentation doesn't say. It's likely necessary to use `-D` argument.

I think I may have another solution for this. By continue use `plat`, except make default string to `runner.os` value. Then override it base on generator string.

Margen67: > 
> 
> Can we rename runner.os to another name such as Android or Xbox One? 😉 I don't think GitHub Actions will have such feature. I do know CMake do have ability compile for other platforms. How? I'm not sure since their generator documentation doesn't say. It's likely necessary to use `-D` argument.
> 
> I think I may have another solution for this. By continue use `plat`, except make default string to `runner.os` value. Then override it base on generator string.

You can continue using `plat` if other things are planned to be supported, then.

RadWolfie: After thinking about the expression method GitHub provide, I don't think it is possible since expression already resolve into a string from the key's variable. Expression within expression wouldn't work either. Which will also cause issue.

In summary, using `runner.os` wouldn't work for default value since runner haven't run yet which make sense. `defaults` currently only support `run` at the moment.

PR #113, opened 2020-09-07
XONLINE: Add comment of titles contain full library to xonline file

RadWolfie:
This is only information add to the comment section.


RadWolfie:
Since it's only additional info to the list, merging it.

PR #117, opened 2020-10-05
Fix D3DDevice::SetTexture and D3DDevice::SetStreamSource LTCG variation for xdk-3911

CookiePLMonster:
This variation puts Stage on the stack and pTexture in `eax`. The existing patch was shared with xdk-4721 with a wrong assumption that both functions are the same, but they were not.

Test case: NASCAR Heat 2002


CookiePLMonster:
I will have more fixes for xdk-3911 D3D8LTCG coming (at least two more functions with incorrect definitions and correct signatures), so it might make sense to hold this PR off until I got those implemented. Marking as draft.

src/OOVPADatabase/D3D8LTCG/3911.inl
```diff
@@ -905,7 +905,7 @@ OOVPA_END;
 // * D3DDevice_SetTexture
 // ******************************************************************
 //81C10000F8FF89 ...C20400
-OOVPA_NO_XREF(D3DDevice_SetTexture_4, 2024, 9)
+OOVPA_NO_XREF(D3DDevice_SetTexture_0, 2024, 9)
```
RadWolfie, line 908: Here's what I suggest to change.

```suggestion
OOVPA_NO_XREF(D3DDevice_SetTexture_4__LTCG_eax_pTexture, 2024, 9)
```

Since non-LTCG version is actually 2 "push", 2\*4. And this call only did one push, 1\*4, and move pTexture to eax register.

src/OOVPADatabase/D3D8LTCG/3911.inl
```diff
@@ -905,7 +905,7 @@ OOVPA_END;
...
 
         { 0x00, 0x83 },
         { 0x01, 0xEC },
```
RadWolfie, line 911: I suggest you to find where eax is set to a register or a location. Then add offset value pair(s) in here to reduce false positive.
i.e.
```
OV_MATCH(0x00, 0x83, 0xEC),
OV_MATCH(0x27, 0x81, 0xC1, 0x00, 0x00, 0xF8, 0xFF, 0x89),
```

RadWolfie (changes requested):

src/OOVPADatabase/D3D8LTCG/3911.inl
CookiePLMonster: Done, matched on `mov esi, eax`.

CookiePLMonster:
Updated with an updated signature for xdk-3911 LTCG version of `SetStreamSource`. This LTCG version passes the first parameter in EDX, but the existing signature matched it to a non-LTCG version instead.

src/OOVPADatabase/D3D8LTCG/3911.inl
```diff
@@ -905,18 +905,11 @@ OOVPA_END;
 // * D3DDevice_SetTexture
 // ******************************************************************
 //81C10000F8FF89 ...C20400
-OOVPA_NO_XREF(D3DDevice_SetTexture_4, 2024, 9)
+OOVPA_NO_XREF(D3DDevice_SetTexture_4__LTCG_eax_pTexture, 2024, 9)
```
RadWolfie, line 908:
```suggestion
OOVPA_NO_XREF(D3DDevice_SetTexture_4__LTCG_eax_pTexture, 2024, 11)
```
There are total of 11 offset value pairs, eventually this manual count will be remove in the future.

src/OOVPADatabase/D3D8LTCG/3911.inl
```diff
@@ -905,18 +905,11 @@ OOVPA_END;
...
-        { 0x27, 0x81 },
-        { 0x28, 0xC1 },
-        { 0x29, 0x00 },
-        { 0x2A, 0x00 },
-        { 0x2B, 0xF8 },
-        { 0x2C, 0xFF },
-        { 0x2D, 0x89 },
+        OV_MATCH(0x00, 0x83, 0xEC), // sub esp, ?
+        OV_MATCH(0x1A, 0x8B, 0xF0), // mov esi, eax
+        OV_MATCH(0x27, 0x81, 0xC1, 0x00, 0x00, 0xF8, 0xFF, 0x89), // add ecx, 0FFF80000h / mov ?
```
RadWolfie, line 912:
```suggestion
        // sub esp, ...
        OV_MATCH(0x00, 0x83, 0xEC),

        // mov esi, eax
        OV_MATCH(0x1A, 0x8B, 0xF0),

        // add ecx, 0xFFF80000; mov ...
        OV_MATCH(0x27, 0x81, 0xC1, 0x00, 0x00, 0xF8, 0xFF, 0x89),
```
See reference from [wiki page of how to write it](https://github.com/Cxbx-Reloaded/XbSymbolDatabase/wiki/Maintaining-OOVPAs-for-Symbol-detection#too-long-didnt-read). Wiki's example of symbol + offset in the comment section is not a requirement.

RadWolfie (changes requested):
Final last two changes, then it will be approved.

RadWolfie (approved):

PR #118, opened 2020-10-07
Add a signature for D3DDevice_DrawIndexedVertices for xdk-3911 LTCG

CookiePLMonster:
Practically enables rendering in xdk-3911 LTCG games, in this case NASCAR Heat 2002.

Marking as draft in case I find more required patches tomorrow or on Friday, but if I don't I'll just mark it as complete.

### DIFF
--- a/src/OOVPADatabase/D3D8LTCG/3911.inl
+++ b/src/OOVPADatabase/D3D8LTCG/3911.inl
@@ -696,6 +696,24 @@ OOVPA_NO_XREF(D3DDevice_DrawIndexedVerticesUP, 1024, 10)
 OOVPA_END;
 
 // ******************************************************************
+// * D3DDevice_DrawIndexedVertices
+// ******************************************************************
+OOVPA_NO_XREF(D3DDevice_DrawIndexedVertices, 1024, 15)
+
+        // push ebp
+        OV_MATCH(0x00, 0x55),
+
+        // mov eax, [esi+0x0478]
+        OV_MATCH(0x0D, 0x8B, 0x86, 0x78, 0x04, 0x00, 0x00),
+
+        // mov dword ptr [eax], 000417FC
+        OV_MATCH(0x51, 0xC7, 0x00, 0xFC, 0x17, 0x04, 0x00),
+
+        // lea ebx, [ebx+0]
+        OV_MATCH(0x15A, 0x8D, 0x9B),
+OOVPA_END;
+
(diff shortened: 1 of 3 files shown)


CookiePLMonster:
Since my current test case has pretty much perfect visuals for now, I doubt I'll find anything more anytime soon. Marking as ready.

src/OOVPADatabase/D3D8LTCG/3911.inl
```diff
@@ -695,6 +695,24 @@ OOVPA_NO_XREF(D3DDevice_DrawIndexedVerticesUP, 1024, 10)
...
+OOVPA_NO_XREF(D3DDevice_DrawIndexedVertices, 1024, 15)
+
+        // push ebp
+        OV_MATCH(0x00, 0x55),
+
+        // mov eax, [esi+0x0478]
+        OV_MATCH(0x0D, 0x8B, 0x86, 0x78, 0x04, 0x00, 0x00),
+
+        // mov dword ptr [eax], 000417FC
+        OV_MATCH(0x51, 0xC7, 0x00, 0xFC, 0x17, 0x04, 0x00),
```
RadWolfie, line 710: I notice 0x000417FC address is relying on "game" specific detection. We do not use hardcode address in any of our database in order to have better detection across titles.

The reason we do not allow this since xbe's sections can be resize at any time, depending on what APIs are used. Or shift around, such as add/remove sections. No xbes are exactly the same nor sort in same ordering.

RadWolfie (changes requested):

src/OOVPADatabase/D3D8LTCG/3911.inl
CookiePLMonster: I based this off the respective signature for a non-LTCG version:
https://github.com/Cxbx-Reloaded/XbSymbolDatabase/blob/master/src/OOVPADatabase/D3D8/3911.inl#L1742

So either this is not an address or the other signature should be corrected too.

jarupxx: I check 0x000417FC address is API specific detection. So we do approve this address.

RadWolfie: My apologies, I should had check from my end first. It is not an actual address reference. 👍

RadWolfie (approved):
LTGM

PR #120, opened 2020-10-21
Fix Kernel Thunk Table Not Being Found Outside of Section Filter List

RadWolfie:
Simple test case is NFL Blitz 2002 title which has kernel thunk table within **.idata** section. It is not on the filter list, therefore require to enforce add section outside of filter list.

Plus add chance to avoid false positive if section doesn't have preload flag just in case. And also used for optimization too.

### DIFF
--- a/src/lib/libXbSymbolDatabase.c
+++ b/src/lib/libXbSymbolDatabase.c
@@ -633,37 +633,54 @@ uint32_t XbSymbolDatabase_GenerateSectionFilter(const void* xb_header_addr, XbSD
     unsigned int section_index = 0;
     unsigned int count = 0;
     const char* SectionName;
+    uint32_t kernel_thunk_addr;
 
     if (pXbeHeader->pSectionHeadersAddr != 0) {
+
+        kernel_thunk_addr = XbSymbolDatabase_GetKernelThunkAddress(xb_header_addr);
 
         for (section_index; section_index < section_total; section_index++) {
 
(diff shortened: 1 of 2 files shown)


src/lib/libXbSymbolDatabase.c
```diff
@@ -665,6 +671,14 @@ uint32_t XbSymbolDatabase_GenerateSectionFilter(const void* xb_header_addr, XbSD
                     break;
                 }
             }
+
+            // If section is not on the filter list, check if there is kernel thunk table within section to enforce include it.
+            if (count == count_check) {
+                if (kernel_thunk_addr >= sh_index->dwVirtualAddr && kernel_thunk_addr < (sh_index->dwVirtualAddr + sh_index->dwSizeofRaw)) {
+                    // Once found within range, then enforce add section to the filter.
+                    goto enforceAddToSectionFilter;
```
PatrickvL, line 679: The use of goto is generally considered a bad code smell; It's probably better to extract-method the entire block that needs to be performed (again), any required variables passed in as arguments.

PatrickvL (changes requested):

src/lib/libXbSymbolDatabase.c
RadWolfie: We do have goto in our codebase. However, it is a necessary to perform clean up properly. Now, in this pull request no longer have goto. 😃

RadWolfie:
It's ready for review/merge. 🙂

PatrickvL (approved):

PR #121, opened 2020-10-30
Direct3D LTCG OOVPA improvements

CookiePLMonster:
* Fixes a ` D3D_BlockOnTime` signature - makes it a bit more generic so the signature matches properly on xdk-5223. Test case: **Midtown Madness 3**
* Splits ` Direct3D_CreateDevice_16` into two separate signatures, as the current OOVPA was matching the same symbol for two variations of `CreateDevice` using a slightly different calling convention:
  * xdk-3911 - passes 4 arguments on stack + in EAX and EBX registers. Test case: **NASCAR Heat 2002**
  * xdk-4627 - passes 4 arguments on stack + in EAX and ECX registers. Test case: **Aggressive Inline**
  * xdk-5028 - passes 4 arguments on stack + in EAX and ECX registers. Test case: **Midtown Madness 3**


RadWolfie:
And what exactly reside on eax and ebx/ecx registers?
See past pull request you made of how you know what they are.
* PR https://github.com/Cxbx-Reloaded/XbSymbolDatabase/pull/117

---

**SIDE NOTE:** Direct3D_CreateDevice_16 signatures (unchanged in this pull request) are still weak and should be improve to lock down which instruction use those registers. It can have greater chance of false positive. Since it is unchanged, it can be improve in later pull request.

CookiePLMonster:
The parameters are `BehaviorFlags` and `ppReturnedDeviceInterface`, are you sure we want to make those function names **this** long? Ultimately it doesn't matter for signatures *what* those parameters are, only *where* they are.

RadWolfie:
For LTCG signatures? Yes, input as eax and ebx/ecx can be confusing what exactly reside on those registers. Heck, compilers can even swap two arguments from the registers if it want to.

CookiePLMonster:
I don't disagree, but like I said it ultimately doesn't matter *what* is in those registers - currently we convey the information on *where* the data is, and for signatures that should be enough.

RadWolfie:
🤷 Without any documentation in the repo, it's not going to be merge. And it does matter, otherwise we could be doing things wrong from third-party's end. No matter how long it can be, that's how it's going to roll.

CookiePLMonster:
Fine - changed.

gandalfthewhite19890404:
I have found **Pro Racer Driving** is recognized as Direct3D_CreateDevice_16__LTCG_eax_BehaviorFlags_ecx_ppReturnedDeviceInterface. 
D3D8LTCG XDK is 5233

RadWolfie (approved):

PR #123, opened 2020-11-05
DSound: Split CDirectSoundStream_GetStatus into Two Symbols for Difference Handler

RadWolfie:
I already verified titles with 4039 and 4134 build. CDirectSoundStream_GetStatus's function of 4039 and below only use one flag which check if CMcpxStream class is ready for input stream packets. 4134 function and later use extended feature of general status.

Low risk merge since I will make a new pull request to Cxbx-Reloaded for resolve GetStatus difference between two different revisions.

### DIFF
--- a/src/lib/libXbSymbolDatabase.c
+++ b/src/lib/libXbSymbolDatabase.c
@@ -1892,8 +1892,14 @@ static bool manual_scan_section_dsound(iXbSymbolContext* pContext,
             internal_RegisterSymbol(pContext, pLibrarySession, XRefNoSaveIndex, 3911,
                 "CDirectSoundStream_GetInfo", *(uint32_t*)(pFuncAddr + 2 * 4));
 
-            internal_RegisterSymbol(pContext, pLibrarySession, XRefNoSaveIndex, 3911,
-                "CDirectSoundStream_GetStatus", *(uint32_t*)(pFuncAddr + 3 * 4));
+            if (pLibrary->build_version < 4134) {
+                internal_RegisterSymbol(pContext, pLibrarySession, XRefNoSaveIndex, 3911,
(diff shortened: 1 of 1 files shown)


src/lib/libXbSymbolDatabase.c
```diff
@@ -1892,8 +1892,14 @@ static bool manual_scan_section_dsound(iXbSymbolContext* pContext,
             internal_RegisterSymbol(pContext, pLibrarySession, XRefNoSaveIndex, 3911,
                 "CDirectSoundStream_GetInfo", *(uint32_t*)(pFuncAddr + 2 * 4));
 
-            internal_RegisterSymbol(pContext, pLibrarySession, XRefNoSaveIndex, 3911,
-                "CDirectSoundStream_GetStatus", *(uint32_t*)(pFuncAddr + 3 * 4));
+            if (pLibrary->build_version < 4134) {
+                internal_RegisterSymbol(pContext, pLibrarySession, XRefNoSaveIndex, 3911,
+                    "CDirectSoundStream_GetStatus__r1", *(uint32_t*)(pFuncAddr + 3 * 4));
```
CookiePLMonster, line 1897: For documentation purposes, maybe it's better to name those symbols by XDK revision of when they first appeared? So `r3911` for this one and `r4134` for the newer one.

RadWolfie: What is the purpose of 3911 and 4123 numbers you see before the symbol's string name? 😉 If you have use Cxbx-Reloaded, you don't see any XDK build versions in symbol cache files. However, if you set kernel log to console or file from Cxbx-Reloaded, you will see XDK build versions return on first run. Therefore is documented.

In summary, we don't need redundant numbers for each API change. We only need to recognize the function's output has change or the inputs as well. Then apply appropriate expectation for inputs and outputs.

Issue #112, opened 2020-09-06
Feature: Add Monthly Release Support


RadWolfie:
What I am looking for is have ability to manually update major and minor numbers from a file. However, patch number will follow total of git commits and if it's from a pull request then append pr-#### for difference in artifact storage. At least have GitHub Actions workflow handle routine releases by monthly if possible and by creating a tag manually.

Chances of have all of features may will require create an action to do the tasks.

PR #125, opened 2020-12-18
Let CI generate releases

thrimbor:
This is pretty much the same thing that is done on the Cxbx-Reloaded repo, having a CI job that just downloads the generated artifact and creates a pre-release, so that it can be download by other workflows or people not logged into GH.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,3 +83,38 @@ jobs:
             README.md
             LICENSE
           if-no-files-found: error
+  release:
+    if: |
+      github.event.action != 'pull_request' &&
+      github.ref == 'refs/heads/master' &&
+      github.repository == 'Cxbx-Reloaded/XbSymbolDatabase'
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Download artifact
+        uses: actions/download-artifact@v1
+        with:
+          name: XbSymbolDatabase
+      - name: Prepare artifact for release
+        run: |
+          echo "short_commit_sha=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
+          7z a -mx1 "XbSymbolDatabase.zip" "./XbSymbolDatabase/*"
+      - name: Create release
+        id: create_release
+        uses: actions/create-release@v1.1.1
+        env:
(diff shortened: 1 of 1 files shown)


RadWolfie:
I have no intention to make nightly build releases. They can be obtain from temporary artifact's end or build it locally. See https://github.com/Cxbx-Reloaded/XbSymbolDatabase/issues/112 for details of plan to use release system purpose.

External GitHub's CI, outside of this repo, can obtain artifact file by the way.

mborgerson:
Hi @RadWolfie, 

> External GitHub's CI, outside of this repo, can obtain artifact file by the way.

If you are aware of how to do this easily, can you please explain. In my experience, having a rolling GitHub release created on 'push to master' events makes the most sense here, especially since all of the necessary artifacts are already created in the process of the CI cycles on this repo. Doing so makes downloading the latest release as simple as `wget https://github.com/mborgerson/XbSymbolDatabase/releases/latest/download/XbSymbolDatabaseTool.zip`, whereas with artifacts one must be logged in to GitHub and figure out how to fetch the URL of the most recent artifact for a branch.

I hope you'll reconsider merging this PR to lower the unnecessary burden on people trying to integrate this tool in other workflows (e.g. my ghidra-xbe plugin).

RadWolfie:
There is API for getting a list of artifacts to choose which artifact to get. https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#artifacts.

@JayFoxRox made a python script for this which can be found at https://github.com/JayFoxRox/GitHub-artifact-URL/.

As for simplify integration with GitHub workflow, there is ["download latest artifact from external GitHub workflow"](https://github.com/marketplace/actions/download-latest-artifact-from-a-github-workflow) action.

'push to master' events are simply nasty hack to store artifacts permanent and pollute the releases as you can see in Cxbx-Reloaded's repo. There's also a lot of confusions due to GitHub doesn't provide a way to disable "semver" from same day releases. However, Cxbx-Reloaded need it anyway for end-user. Here in XbSymbolDatabase repo is not even necessary. Nor is intended for end-user usage.

For local users building from your repo, they will have to wait until there's monthly ***versioning*** release system is implement. Or they can simply download either monthly stable release or artifact build in respective folder.

PR #139, opened 2021-06-18
Xapi's Memory Unit Sigs Update

RadWolfie:
Add two new signatures and one derive address into database.

As for `g_XapiMountedMUs` symbol to register. It require manual scan of `MU_Init` and `IUsbInit_GetMaxDeviceTypeCount` for dynamic search reference location. Since `MU_Init` do change over many xdk revisions. This is only best way to handle without need to perform extra workload to make individual `MU_Init` sigs for each offset difference.

This work is a joint effort with @ergo720. Depending on what other things may need to be add.


src/lib/libXbSymbolDatabase.c
```diff
@@ -1931,6 +1943,119 @@ static bool manual_scan_section_dsound(iXbSymbolContext* pContext,
...
+        }
+
+        // Look for possible push and call next to each other.
+        if (cur[0] == 0x68 && cur[5] == 0xE8) {
+
+            // check if call is linked to IUsbInit_GetMaxDeviceTypeCount function.
+            xbaddr ActualAddr = *(xbaddr*)(cur + 6);
+            if (MatchXRefAddr(cur + 6, virt_start_relative, pContext->xref_database[XREF_IUsbInit_GetMaxDeviceTypeCount])) {
+                // this is where g_XapiMountedMUs hardcode address reside in.
+                xXbAddr = *(xbaddr*)(cur + 1);
```
ergo720, line 2042: This is wrong. This is not `g_XapiMountedMUs`, but the MU xpp type. Here is how `MU_Init` looks like in the game Phantasy Star Online
Capture
As you can see, the push instruction has the address of an xpp type, not of an ULONG, which is what `g_XapiMountedMUs` is. This global is instead referenced by `XUnmountMU`, and you can check this code [here](https://github.com/ergo720/Cxbx-Reloaded/blob/364fbb610dc6286b1b918cdb117505d0b085963e/src/core/hle/XAPI/Xapi.cpp#L446) to see how I'm deriving it in cxbxr.

ergo720 (changes requested):

src/lib/libXbSymbolDatabase.c
RadWolfie: Thanks, I got confused for a moment when you shown me the offset with `XPP_DEVICE_TYPE` which said otherwise. I made a re-word to `g_DeviceType_MU` and rename the commit. Since `g_XapiMountedMUs` can be easily derive from signatures itself. I only need to manually register xref at end of the scan process.

ergo720 (approved):

PR #140, opened 2021-06-18
Fix CLI Tool missing XbSymbolContext_RegisterXRefs call

RadWolfie:
While testing with Ghidra-Xbe extension with Ghrida. I notice couple symbols were not output, including #139 pull request.

This pull request now do manual output derived references that is not part of OOVPA signature database after second part of the scan process.


RadWolfie:
Low risk merge

PR #141, opened 2021-06-20
Added XReadMUMetaData

ergo720:
This is another function necessary to support MUs in Cxbx-Reloaded.


RadWolfie (approved):
Already verified with the offsets and references for titles with 4831+ builds.

Issue #62, opened 2019-03-21
TASK: Replace default OV array to macro usage


RadWolfie:
Library databases to work on:
* D3D8 **(Assigned to: __ )**
* D3D8LTCG **(Assigned to: __ )**
* DSound **(Do not fix this database)**
* JVS **(Assigned to: __ )**
* XActEng **(Assigned to: __ )**
* Xapi **(Assigned to: __ )**
* XGraphic **(Assigned to: __ )**
* XNet **(Assigned to: __ )**
* XOnline **(Assigned to: __ )**

Using generic offset value pair array are no longer advice to use:
```
{ 0x00, 0x04 },
{ 0x01, 0x56 },
{ 0x10, 0x03 },
...
```

It will need replace to use OV_ENTRY or OV_ENTRIES macro:
```
OV_ENTRIES(0x00, 0x04, 0x56),
OV_ENTRY(0x10, 0x03),
...
```

PR #143, opened 2021-07-04
revise D3DDevice_KickOff oovpa for 5455, fit 5849

jackchentwkh:
revise D3DDevice_KickOff oovpa for 5455
original oovpa won't fit on 5849
the newly revised oovpa is much simpler and fit for 5455~5849
verified with Otogi:5455, 5849 xdk beginpush sample.


src/OOVPADatabase/D3D8/5455.inl
```diff
@@ -23,6 +23,7 @@
 // *
 // ******************************************************************
 
+
```
ergo720, line 26: Unnecessary line added.

src/OOVPADatabase/D3D8/5455.inl
```diff
@@ -272,39 +273,22 @@ OOVPA_NO_XREF(D3DDevice_GetTile, 5455, 12)
...
            XRefZero)
 {
+    { 0x00, 0x56 },
+
+    { 0x01, 0x8B },
+    { 0x02, 0xF1 },
+
+    { 0x03, 0x8B },
+    { 0x04, 0x46 },
+    { 0x05, 0x08 },
```
ergo720, line 288: You should use the `OV_MATCH` macro instead of the old array method. See https://github.com/Cxbx-Reloaded/XbSymbolDatabase/issues/62

PatrickvL: Also, for ease of reading, consider adding an end of line comment mentioning the disassembled opcode and its arguments

src/OOVPADatabase/D3D8/5455.inl
jackchentwkh: removed

src/OOVPADatabase/D3D8/5455.inl
jackchentwkh: revised to OV_MATCH
unable to add opcode comment because OV_MATCH put everything in one line.

PatrickvL: `OV_MATCH(....); // MOVC ..., ...`

PatrickvL: One line per opcode

jackchentwkh: interesting, I search the whole database, seems I am the first one to use OV_MATCH.

jackchentwkh: done.

RadWolfie:
@PatrickvL's quote:
> OV_MATCH(....); // MOVC ..., ...

That's not how we do it here. See https://github.com/Cxbx-Reloaded/XbSymbolDatabase/wiki/Maintaining-OOVPAs-for-Symbol-detection#too-long-didnt-read as example reference and active OV_MATCH in the general codebase.

There are times we have a comment note on same line as OV_MATCH for any difference recongization over time is what reserved for.

RadWolfie:
Base on this signature revision, I fail to see how it can increase strength when it's using common known op code.

RadWolfie:
```c
OOVPA_XREF(D3DDevice_KickOff, 5455, 1 + 12,

           XREF_D3D_CDevice_KickOff,
           XRefOne)
{
    // mov eax, XREF_D3DDEVICE
    XREF_ENTRY(0x20, XREF_D3DDEVICE), // Derived

    // push esi
    // mov esi, ecx
    OV_MATCH(0x00, 0x56, 0x8B, 0xF1),
    // mov eax, [esi + 8]
    // test al, 0x04
    OV_MATCH(0x03, 0x8B, 0x46, 0x08, 0xA8, 0x04),

    // test ah, 0x20
    OV_MATCH(0x14, 0xF6, 0xC4, 0x20),

    // mov eax, XREF_D3DDEVICE
    OV_MATCH(0x19, 0xA1),

} OOVPA_END;
```
Above signature is suggestive way to safeguard false positives and use proper format in our current codebase's method.

jackchentwkh:
@RadWolfie I read your OOVPA, it's very clear and also suit 5849.
but when I use it and run with xdk beginpush sample, it won't detect the kickoff() as expected.
I have to move the 
XREF_ENTRY(0x20, XREF_D3DDEVICE), // Derived
to make it work.
it's very strange. 
the XREF_D3DDEVICE is indeed detected as Symbol: 0x0002f308 -> D3DDEVICE 0 which does match the disassembly code.
anyone can explain what might be wrong?

RadWolfie:
Actually, that was my fault. It should be offset 0x1A for XREF_D3DDEVICE address. I didn't use the hexadecimal increment process in my mind.

jackchentwkh:
Haa, I made the same mistake when I checked the code with the disassembly.
thanks a lot.

RadWolfie (approved):
Already verified titles with 5455, 5788, and 5849 builds.

PR #151, opened 2021-11-04
Feature: Full XReference Cache System

### LINKED ISSUE #87: Search Process is Unoptimized
Base on @JayFoxRox's feedback about XbSymbolDatabase. The search process is unoptimized because of not using better algorithms to gain faster scan process.

> Searching for XInput for example, shouldn't take longer than like.. 100ms [but it took like 10 seconds on original Xbox, when I last tried afair - which is too long]

Currently, the scan process check for OV then move on to next offset, +1. If I recall correctly, it also check the reference address first.

* https://en.wikipedia.org/wiki/Knuth%E2%80%93Morris%E2%80%93Pratt_algorithm

In my case, it is D3D8 library taking way much longer time in `.text` section than others.

RadWolfie:
Currently in draft at the moment, reasons:
1. Need to migrate various altered symbol names into single xref usage. (will be done in separate commit for readability)
   * This currently affect unnamed cross reference symbols from revisions.
2. At least ability to determine non-LTCG and LTCG xrefs need to remain separate for X reason.
3. Need to purge global cross reference strings list.

---

With full xreference cache system, we can reduce unnecessary rescan same symbol. Except, we do still have other optimization to do. I believe this pull request partially resolve #87 ticket.

With cross reference's symbols separated into their own library file. It will be easier to maintenance and organizing.

Pros:
1. Allow context's xref_database to allocate and cache for every linked symbols.
2. Reduce partial of redundant scan process system and stop re-report for same address.

Cons:
1. Add new symbol, aka group or derived usage, will require manual add into specific library's def file.
   * Compiler will stop building if cross reference symbol doesn't exist. This helps enforce contributor to verify their work. Then add new symbol if it's actually a new symbol.
2. Any symbol rename will also need to rename inside .def file. Using IDE's rename process may will detect other relative same name. Such as in the comment section.


RadWolfie:
Tested and verified no breakage against master branch.

Issue #136, opened 2021-05-25
XInputSetLightgunCalibration is missing


ergo720:
This xinput api is necessary to emulate lightguns but currently, XbSymbolDatabase has no support for it at all. A possible test case for the api is Phantasy Star Online because, when inspecting the symbol cache generated by cxbx-r for it, I can see that `XInputGetDeviceDescription` is detected, which is another api required for lightguns. Thus, it's likely that the title also has the missing api as well.

ergo720:
It turns out that Phantasy Star Online does not have the LIGHTGUN library and thus it doesn't have `XInputSetLightgunCalibration`.

PR #152, opened 2021-11-07
Added XInputSetLightgunCalibration

ergo720:
Closes https://github.com/Cxbx-Reloaded/XbSymbolDatabase/issues/136. I found this function in the games House of the dead 3 (xbox version) and Starsky & Hutch in the "LIGHTGUN" library with xdk version 4831 and 5344 respectively, The OOVPA is able to detect both xdk versions, but I don't know if it will work with older and newer xdks since I don't have other test games for those.


src/OOVPADatabase/Xapi/4831.inl
```diff
@@ -206,6 +206,32 @@ OOVPA_SIG_MATCH(
     //
 );
 
+// ******************************************************************
+// * XInputSetLightgunCalibration
+// ******************************************************************
+OOVPA_SIG_HEADER_XREF(XInputSetLightgunCalibration,
+                      4831,
+
+                      XRefNoSaveIndex,
```
RadWolfie, line 215: With full xreference cache system pull request merged, these lines should be removed.
```suggestion
```

src/OOVPADatabase/Xapi/4831.inl
```diff
@@ -206,6 +206,32 @@ OOVPA_SIG_MATCH(
...
+                      XRefOne)
+OOVPA_SIG_MATCH(
+
+    XREF_ENTRY(0x1C, XREF_KT_FUNC_KeRaiseIrqlToDpcLevel),
```
RadWolfie, line 219: Can you add a comment what exactly this entry do? Call? Jump to?

Also, perhaps add the instruction's bytes before this entry for safeguard purpose? (just a nit)

RadWolfie (changes requested):
Two review remarks and a rebase request.

Then everything looks good to me.

RadWolfie (approved):

PR #157, opened 2022-06-18
Enable XOnline symbol detection when XONLINLS library is found

thrimbor:
This enables detection of XOnline symbols when the `XONLINLS` library is found. I don't know how exactly it differs from `XONLINES`, but the current signatures seem to be enough for symbol detection to work. The library version seems to always be `5849`, and it can be found in games such as `TOCA Race Driver 2`, `Trivial Pursuit` and `RalliSport Challenge 2`.

### DIFF
--- a/include/libXbSymbolDatabase.h
+++ b/include/libXbSymbolDatabase.h
@@ -70,6 +70,7 @@ extern "C" {
 #define Lib_XONLINE  "XONLINE"
 #define Sec_XONLINE  Lib_XONLINE
 #define Lib_XONLINES "XONLINES"
+#define Lib_XONLINLS "XONLINLS"
 #define Sec_XNET     "XNET"
 
 #define XbSymbolLib_D3D8     (1 << 0)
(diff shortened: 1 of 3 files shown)


RadWolfie:
Thanks for the finding! Just a small favor from your end, can you force-push your branch in order to trigger the CI builds please?

thrimbor:
Done, I'm not sure what caused it to not run them before.
The build failures on Windows seem to be caused by VS 2019 not being there, the current Windows image seems to have VS 2022 installed.

RadWolfie:
> Done, I'm not sure what caused it to not run them before.

What caused them not to run before is CI workflow has job scheduler for specific date to perform a release. When repository has no activity for at least 2 months, GitHub will automatically disable workflow containing job scheduler task. I will correct this problem by split CI workflows to only have one workflow has scheduler and other mainly for CI workflow. 

> The build failures on Windows seem to be caused by VS 2019 not being there, the current Windows image seems to have VS 2022 installed.

Ah, that's right. I forgot about this. I'll need to remove VS build version target and let it run whichever CMake can find.

RadWolfie (approved):
After carefully review of additional library placement, I can confirm all required locations are set for proper functional scan process.

PR #158, opened 2022-06-21
CI: Update to split scheduled release into separate workflow

RadWolfie:
With this workflows update, CI workflow will no longer automatically disabled after repository's inactivity for over two or three months.

We can manually activate Schedule Release workflow and run at any time without being affected from pull request's CI builds.

Plus updated CI workflow to exclude Visual Studio fixed generator to allow CMake to find current Visual Studio generator.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,12 @@
 name: GitHub CI
 
 on:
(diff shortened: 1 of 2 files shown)


RadWolfie:
I have verified release schedule is working as intended.

PR #164, opened 2022-09-22
XAPI: Add two symbols to complete XMU prefix set

RadWolfie:
Complete a set of XMU prefix symbols from todo list.

Added:
- XMUPortFromDriveLetterA (3911)
- XMUSlotFromDriveLetterA (3911)

Had been tested with titles from 3911 to 5849 builds with no issue.


src/OOVPADatabase/Xapi/3911.inl
```diff
@@ -1893,3 +1895,51 @@ OOVPA_SIG_MATCH(
...
+// * XMUPortFromDriveLetterA
+// ******************************************************************
+// Generic OOVPA as of 3911 and newer.
+OOVPA_SIG_HEADER_NO_XREF(XMUPortFromDriveLetterA,
+                         3911)
+OOVPA_SIG_MATCH(
+
+    // mov al, byte [esp + 4]
+    OV_MATCH(0x00, 0x8A, 0x44, 0x24, 0x04),
+    // cmp al, 0x46
```
ergo720, line 1909: Nit: missing new line between the `OV_MATCH` and the comment. Same goes for `XMUSlotFromDriveLetterA` below.

RadWolfie: If you count the bytes to match which will sum up to 4. Next offset is 4 which is what you see on next line. So, both instructions reside next to each other. Only time they need separate line if we skip amount of instructions.

Here's an example:
https://github.com/Cxbx-Reloaded/XbSymbolDatabase/blob/61615046d91bf8d68491238be96eff5a6fdc4b47/src/OOVPADatabase/D3D8LTCG/5344.inl#L30-L49

ergo720: In that case, this convention is not enforced consistently in the code base. Take for example `D3DDevice__m_VerticalBlankEvent__ManualFindGeneric` xdk 3911. There is a two byte `OV_MATCH` at offset 0x15, followed by another one at offset 0x17, and yet it still skips a new line. In my opinion, not skipping a new line between ov matches just looks ugly and doesn't add any useful information, so you should always skip a new line regardless.

RadWolfie: `D3DDevice__m_VerticalBlankEvent__ManualFindGeneric`'s signature is my mistake for old format and missed instructions next to each other. However, it is 3 years old signature which need a re-format again. Your opinion is noted, but it is useful for readable format. Having each (partial) instructions on every new line separator simply waste of space unless they don't reside to each other. I explained very basic detail which I will quote again.
> So, both instructions reside next to each other. Only time they need separate line if we skip amount of instructions.

Which translate in "disassembly" readable view.
```asm
mov ecx,[eax + -0x1] 
dec eax 
dec ecx 
mov [eax],ecx 
JNZ +0xB
...
call XMemFree
...
```

ergo720: Alright, fine by me. However, I suggest to at least document this "skipping line" logic somewhere here in XbSymbolDatabase (for example, in the "OOVPA Signature Guideline" page of the wiki), so that users don't just assume that the `OV_MATCH` is not properly formatted.

ergo720 (approved):

PR #165, opened 2022-09-25
XAPI: Add Device types Manual Scan

RadWolfie:
Migrate Cxbx-Reloaded's [`SetupXboxDeviceTypes`](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/42046400524cb837c2290c0bc008226592455d74/src/core/hle/XAPI/Xapi.cpp#L400-L488) function into `internal_xapi_find_device_types` function.

With this changes, we are able to scan and label each known global device type variables. The following device types supported are:
- Gamepad (ucType=1)
- Keyboard (ucType=2)
- IR Dongle (ucType=3)
- Steel Battalion Controller (ucType=128)

With verification for XInputOpen function to find device types. Titles compiled below 4242 build are provided with static device types. Any titles compiled as of 4242 and later build has ability to add/remove any device types on compile time. Plus ability include `g_DeviceTypeInfoTable(Begin|End)` variables to perform your own lookup method.

**NOTE:  `g_DeviceTypeInfoTable(Begin|End)` is only available for titles compiled from 4242+ build.**

`internal_section_VirtToHostAddress`'s helper function is made easier to lookup through each loaded sections in order to find a section contain within virtual address range. This is needed for NFL Blitz 2002 title as there are two external sections, aka ".XID" and ".XPP&Dat", contain device table and types.

I also made `internal_section_HostToVirtAddress` function can be used as well. Currently, it is not in use.


src/lib/manual_xapilib.h
```diff
@@ -149,16 +236,20 @@ static bool manual_scan_section_xapilib(iXbSymbolContext* pContext,
                                   "g_DeviceType_MU", xXbAddr);
     }
 
-    return true;
+    bool foundDeviceTypes = internal_xapi_find_device_types(pContext, pLibrarySession, pLibraryDB, pSection, virt_start_relative);
+
+    return true && foundDeviceTypes;
```
ergo720, line 241: `true &&` is useless and can just be deleted.

include/libXbSymbolDatabase.h
```diff
@@ -61,7 +61,9 @@ extern "C" {
 #define Lib_XACTENG  "XACTENG"
 #define Sec_XACTENG  Lib_XACTENG
 #define Lib_XAPILIB  "XAPILIB"
+#define Sec_XID      ".XID"
 #define Sec_XPP      "XPP"
+#define Sec_XPPDat   ".XPP&Dat"
```
ergo720, line 66: If I consult an ASCII chart, I see that "." comes first than "X", so .XID and .XPP are probably in the wrong order

ergo720 (approved):

PR #170, opened 2022-12-01
HOTFIX: Fix inaccurate symbols comparison error and migrate equal function check

RadWolfie:
This bug was outstanding hidden from plain view for long time. With this fix, the errors of mismatch should now correctly output and do not hide even size errors not being output.


RadWolfie:
I had manually added several variant of tests against new comparison locally and show they are outputting correctly.

PR #171, opened 2022-12-04
Implement Dependency API (removed Mutex callback API)

### LINKED ISSUE #168: Feature: Create priority/dependency API on filtered libraries
To remove mutex callback requirement will need new APIs to allow external projects make decision which will have higher priority over other library in multi-threading usage.

Currently, internal API only perform check which library types are in used and prevent other thread attempt to scan other library with same type of category. For an example, XACTENG and DSOUND libraries should not be scanning at the same time. Reason? XACTENG library have dependency on DSOUND's symbols needed in order to complete the scan process. The same for XONLINE library have dependency on XNET's symbols as another example.

RadWolfie:
With Dependency API feature, there is no more of random occur failures for any partial library's symbols. Dependency API is require for using `XbSymbolContext_ScanLibrary` API.

I am requesting reviews on the API naming or even discovery of faulty code.

I have personally tested this new change on single thread and multi-threads and confirm random failures of missing symbols has stopped except for one weird bug not relative to this pull request.

Weird bug report is: (fixed in commit 3eb125546e9d7c195aef690b1d96a7258fddd22e)
`ERROR  : g_SymbolAddresses(Raw) is missing UNKNOWN (b0000) 0x00000000 ->`

fixed #168

### DIFF
--- a/include/libXbSymbolDatabase.h
+++ b/include/libXbSymbolDatabase.h
@@ -265,28 +265,6 @@ void XbSymbolContext_SetContinuousSigScan(XbSymbolContextHandle pHandle, bool en
 void XbSymbolContext_SetFirstDetectAddressOnly(XbSymbolContextHandle pHandle, bool enable);
 
 /// <summary>
-/// To register mutex lock callback functions.
-/// </summary>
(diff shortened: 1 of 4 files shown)


RadWolfie:
Since there's no feedback from the team, merging as-is.

PR #176, opened 2022-12-27
Forward all std::cout messages to a function

RadWolfie:
Plus add thread id for every message outputs if using multi-threading mode.

Only thing left intact are new lines from std::cout which are needed to split sections of messages from a console/log.


RadWolfie:
Oops! I forgot to include verbose filter for unit test's messages.

PR #193, opened 2023-03-06
Another D3D8(LTCG) signatures update

RadWolfie:
See commits' body for what been changed.

Fixed:
- Regression in Grand Theft Auto: San Andreas for missing `D3DDevice_DrawIndexedVertices` symbol due to small portion of instructions was shifted by one offset.
- False positive for `D3DDevice_DrawVerticesUP` symbol matching with `D3DDevice_DrawIndexedVerticesUP` address.
- Some duplicate signature warnings during scan process and additional detection too.

Renamed some LTCG symbols to align with new change relative to register parameters usage.

Currently unit test is passing after rename correction for both LTCG and non-LTCG titles.


PatrickvL (approved):
Great work!

PR #195, opened 2023-05-08
Change Licensing to MIT

### LINKED ISSUE #72: Relicense
XbSymbolDatabase should be relicensed to CC0 / BSD / MIT or something similar.

It's currently GPL, but that has a couple of drawbacks:

- Some tools, such as IDA Pro are not GPL and we couldn't legally make plugins for that.
- Most appstores don't allow Open-Source (so we need abilities to dual-license), so our own tools / debuggers for smartphones would possibly be impossible to release in the future.
- On Xbox, there's no dynamic linker, so even LGPL wouldn't be too helpful.

So while it's good to force users to provide code open-source, it's also preventing some use-cases or makes distribution more cumbersome.

I think it's bad for this library to force open-source anyway, as it's largely incompatible with its use cases. I can't think of a bad context either, where a non-open-source license would damage this project.

RadWolfie:
close #72

All authors have confirmed permission to use MIT or BSD-2-clause licenses. Since MIT license is earliest flexible license, I have re-licensed all GPL2 or later to their respective licenses.

The changes also include PVS Studio comment update and Open Data Commons Open Database License v1.0 correction in signature's database.


PatrickvL (approved):

ergo720 (approved):

RadWolfie:
Closing pull request due to one contributor permission was missed.

PR #198, opened 2023-06-26
XAPI: Add event signatures

RadWolfie:
The following event relative function signatures are added:
- CreateEventA
- OpenEventA
- PulseEvent
- ResetEvent
- SetEvent

I can confirm they do not change over time at all.


RadWolfie:
Added space after comma in comment requested by @ergo720.

ergo720 (approved):

PR #201, opened 2023-10-28
HOTFIX: fix endless loop for dependency flag check is completed or not

### LINKED ISSUE #197: Titles with D3D8I and D3D8X libraries cause scan process to stuck looping
This issue only happen if D3D8X library exist in database but not D3D8I library which is actually a dependency for D3D8X library.

A better method is needed to determine if dependency library do exist or not to avoid getting stuck waiting on non-existence dependency library to complete.

RadWolfie:
After reviewing codebase, I believe I found the source of endless loop for when dependency flags are returned but never checked if any of them are matched. Instead of fix cli and unittest tools, I recommended to fix it internally to work as intentional design. This fixed certain Chihiro titles containing D3D8I library except absent of D3D8I signatures support.

fixed #197

### DIFF
--- a/src/lib/libXbSymbolDatabase.c
+++ b/src/lib/libXbSymbolDatabase.c
@@ -463,8 +463,8 @@ uint32_t XbSymbolDatabase_GetLibraryDependencies(uint32_t library_flag, XbSDBLib
             ret_dependencies |= library_filters.filters[i].flag;
         }
     }
-    // If flag dependency is/are found, then return those.
-    if (ret_dependencies) {
+    // If library filters were given, then return any dependencies that were found.
+    if (library_filters.count) {
         return ret_dependencies;
     }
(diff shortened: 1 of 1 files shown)


src/lib/libXbSymbolDatabase.c
```diff
@@ -463,8 +463,8 @@ uint32_t XbSymbolDatabase_GetLibraryDependencies(uint32_t library_flag, XbSDBLib
             ret_dependencies |= library_filters.filters[i].flag;
         }
     }
-    // If flag dependency is/are found, then return those.
-    if (ret_dependencies) {
+    // If library filters were given then return what dependencies are found or not.
```
ergo720, line 466:
```suggestion
    // If library filters were given, then return any dependencies that were found.
```

ergo720 (approved):

PR #154, opened 2021-11-10
Extend OOVPA Tests

RadWolfie:
Review remarks are currently welcome at this time and any new internal tests as well.

Reasons this pull request is draft:
- [ ] What I said on first line, mostly for any new internal tests to add.
- [ ] I would like to resolve old errors first before proceed to investigate any further to clean up new errors listed below.
   - So, this pull request will be remain open longer than usual.

---

List of self-tests added:
- SymbolDatabaseVerifyContext_VerifyXRefJmp (as debug message)
  - Intended to be awareness of very minimalize jump instruction to a function. Does not have ability to verify the difference detection. Therefore decided to set message level to debug for verbose.
  - Reason for this was I noticed IDirectSoundStream_SetMixBinVolumes_(8/12) indeed require separate cross reference to ensure proper symbol name is used.
- SymbolDatabaseVerifyContext_VerifySymbolDuplicate (as error message)
  - Avoid using exact same symbol name as possible.
    List of below is pretty much rules should be follow which will be likely copy over to wiki page.
    - If `stdcall` symbol is used, then it require to be normal name unless later API has changed to add / remove parameter.
      - For this, using `_#` suffix is required.
    - If there's some sort of revision changed with same parameters / return. It will be better to use `__r1` suffix or the like. This decision had not been made.
      - `CDirectSoundStream_GetStatus` as an example does not fall under this category. Since the internal function had changed a bit.
    - If symbol is LTCG call type specific, then `__LTCG` suffix is required along with which parameters are assigned to either onto stack or registers. We only need registers assigned, parameters on the stack is not a requirement.
    - More to come later (if anything more to add)
- TBD

## List of current reports as of commit d60c19ab8230a3501ea279bc87ee243c89348af4
SymbolDatabaseVerifyContext_VerifyXRefJmp:
```
DEBUG  : D3DCubeTexture_LockRect has one xref and one jump instruction
DEBUG  : D3DDevice_CreateImageSurface has one xref and one jump instruction
DEBUG  : D3DResource_BlockUntilNotBusy has one xref and one jump instruction
DEBUG  : D3DVolumeTexture_LockBox has one xref and one jump instruction
DEBUG  : CDirectSoundBuffer_SetAllParameters has one xref and one jump instruction
DEBUG  : CDirectSoundBuffer_SetConeAngles has one xref and one jump instruction
DEBUG  : CDirectSoundBuffer_SetConeOutsideVolume has one xref and one jump instruction
[...]

PR #208, opened 2024-07-15
Upgrade OOVPA symbol group files

### LINKED ISSUE #122: Feature: Add Parameter Info into Symbol Group
ergo720 mentioned about ability to store argument info within function's data. While it's under consideration, each symbol group cannot have same name. One of the possibility is to understand in what order they were in non-LTCG symbols and follow that example with LTCG symbols. Such as...

non-LTCG:
`int example_func(void* arg1, unsigned int arg2, short arg3, int arg4)`
LTCG:
`int example_func(unsigned int arg2, int arg4)` **+ arg1 is store in ecx and arg3 is store in dx**.

which then can roughly translate into...

non-LTCG:
`example_func_16`
LTCG:
`example_func_8__LTCG_ecx1_dx3` (as shortest way possible)
Plus document what is the purpose of ecx and dx registers are in the comment section for validation. Eventually, we will need to document their arguments or leave as unknown.

Unless such function doesn't exist in non-LTCG, then we may replace number to `Unk`.

Finally, we could extend and pass these information up to third-party usage for maybe better understanding for both non-LTCG and LTCG. Although, this method will consume some spaces.

However, this is still under consideration and may need further discussion.

---
**NOTE:**
**1) Currently, we only have iterator of every OOVPA signatures into one list per library than per symbol group.**
**2) Will likely use extend function APIs to obtain these information and pass it up to third-party usage.**

RadWolfie:
> [!NOTE]
> (COMPLETED) ~~`upgrade-symbol-groups` branch, in this pull request, is currently in **major work in progress** and not a final design.~~

> [!NOTE]
> ~~GitHub CI build may can fail at any time, do not worry about it.~~

> [!NOTE]
> Some things may not be fully done due to in need for other internal changes to enable such various support.
> Such as self-register symbols support.

resolve #122

Some changes to symbol names are require to functional with new upgrade system.
- Removed XAPI prefix, no longer necessary.
  - ~~Currently undetermined for OFFSET swapped in prefix and suffix field.~~ (now always use suffix)
- D3D8LTCG symbol names have been updated to expected generator symbol name for LTCG version.
  - Register name with reference to parameter name has been optimized to use register name and parameter offset by one.

Also, some symbols may have additional/removed parameters. These symbols will need their own OOVPA revision to separate and provide correct parameters and stack size usage.

New feature for CLI tool:
- [-d -demangle] option will output generic name instead of mangled name.
- [-e -extendinfo] option will output with parameters' name and storage of the symbol; i.e. `DSOUND__FUN__CDirectSoundStream_GetInfo(stk pThis, stk pInfo) = 0x########`.

UnitTest may have not been updated with symbol name changes.

Checklist of symbol group files has been done:
- [x] D3D8
- [x] D3D8LTCG
- [x] DSound
- [x] JVS
- [x] XActEng
- [x] Xapi
- [x] XGraphic
- [x] XNet
- [x] XOnline

Review remarks are welcome. 🙂
Pull request to `upgrade-symbol-groups` branch for incomplete symbol group file are also welcome too.
Best thing to do is look for TODO notes relative to parameters change or the like and any `/*unknown*/` comments.

---
TODO:
- Update self test library function for:
  - [ ] (POSTPONE) Duplicate reference index, list of parameters. (May likely migrate from #154 pull request)
- Update UnitTest tool for:
  - [x] Remove unnecessary numerous of mangled names output if not found reference.

---
**UPDATE:** Calling conventions are added to this pull request due to ghidra-xbe's decompile doesn't understand and couldn't do a manual check for what type of calling conventions for the functions.

ghidra-xbe's future options support is now able to decompile correctly with corrected parameters and calling conventions.


include/libXbSymbolDatabase.h
```diff
@@ -197,15 +232,18 @@ void XbSymbolDatabase_SetOutputMessage(xb_output_message_t message_func);
...
 /// </summary>
 /// <param name="library_str">Name of the library in string.</param>
 /// <param name="library_flag">Name of the library in flag.</param>
 /// <param name="xref_index">Output unique xreference of symbol name.</param>
 /// <param name="symbol_str">Name of the library in symbol string.</param>
 /// <param name="address">Return xbox's virtual address.</param>
-/// <param name="build_verison">Found with specific build verison.</param>
-typedef void (*xb_symbol_register_t)(const char* library_str, uint32_t library_flag, uint32_t xref_index, const char* symbol_str, xbaddr address, uint32_t build_verison);
+/// <param name="build_version">Found with specific build version.</param>
+/// <param name="symbol_type">Type of symbol. If symbol type is a function, then param_ prefix arguments below will be input.</param>
```
ergo720, line 243:
```suggestion
/// <param name="symbol_type">Type of symbol. If symbol type is a function, then the param_ prefix arguments below will be set.</param>
```

include/libXbSymbolDatabase.h
```diff
@@ -223,6 +261,20 @@ bool XbSymbolScan(const void* xb_header_addr, xb_symbol_register_t register_func
 /// <returns>Return "UNKNOWN" string if does not exist in the database. Otherwise will return library name string.</returns>
 const char* XbSymbolDatabase_LibraryToString(uint32_t library_flag);
 
+/// <summary>
+/// To convert parameter type into string format.
+/// </summary>
+/// <param name="param_type">Input provided param_type from symbol register callback.</param>
+/// <returns>Return "unk" string if does not exist in the database. Otherwise will return parameter type string.</returns>
```
ergo720, line 268:
```suggestion
/// <returns>Return "unk" string if does not exist in the database. Otherwise it will return the string representation of the parameter type.</returns>
```

include/libXbSymbolDatabase.h
```diff
@@ -223,6 +261,20 @@ bool XbSymbolScan(const void* xb_header_addr, xb_symbol_register_t register_func
...
+const char* XbSymbolDatabase_ParamToString(uint32_t param_type);
+
+/// <summary>
+/// To convert symbol reference index into string format.
+/// </summary>
+/// <param name="xref_index">Input provided xref_index from symbol register callback.</param>
+/// <returns>Return demangled symbol name string.</returns>
+const char* XbSymbolDatabase_SymbolReferenceToString(uint32_t xref_index);
```
ergo720, line 276:
```suggestion
/// To convert a symbol reference index into string format.
/// </summary>
/// <param name="xref_index">Input provided xref_index from symbol register callback.</param>
/// <returns>Returns a demangled symbol name string.</returns>
const char* XbSymbolDatabase_SymbolReferenceToString(uint32_t xref_index);
```

include/libXbSymbolDatabase.h
```diff
@@ -127,6 +127,41 @@ static const char* xbe_type_str[XB_XBE_TYPE_MAX] = {
...
 };
 
+typedef enum _XbSDSymbolType {
+    symbol_internal = 0, // Only for internal usage.
+    symbol_variable,
+    symbol_function
+} XbSDSymbolType;
+
+typedef enum _XbSDBParamType {
+    param_unk = 0, // Unknown
```
ergo720, line 137: `param_unk` is useless because `XbSymbolDatabase_ParamToString` never checks against it, and instead catches invalid indexes with a `default` keyword. So I suggest this `param_stk = 1,     // 32 bits`

RadWolfie: Please leave this review remark open, I may had given incorrect comment to the API relative to this.

include/libParamTypes.h
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: ODbL-1.0
+
+PARAM_TYPE_32(unk) // Unknown, currently intended to be used internally. May be subject to change accessible to public usage.
```
ergo720, line 3:
```suggestion
PARAM_TYPE_32(unk) // Unknown, currently intended to only be used internally. This may change in the future (e.g. it may become accessible to the user).
```

RadWolfie: See https://github.com/Cxbx-Reloaded/XbSymbolDatabase/pull/208#discussion_r1685670653

Closing as whole comment will be replace with entirely different comment.

include/libXbSymbolDatabase.h
RadWolfie: Resolved in updated fixup commit.

RadWolfie:
Unit tests are passing for all libraries locally.

include/libXbSymbolDatabase.h
```diff
@@ -197,15 +230,18 @@ void XbSymbolDatabase_SetOutputMessage(xb_output_message_t message_func);
 
 /// <summary>
 /// To register any detected symbol name with address and build version back to third-party program.
-/// NOTE: Be aware of library name will be varity since some libraries are detecting in other sections as well.
+/// NOTE: Be aware of library name will be variety since some libraries are detecting in other sections as well.
```
ergo720, line 233:
```suggestion
/// NOTE: Be aware that the library name will vary since some libraries are detected in other sections as well.
```

include/xref/list_xref.h
```diff
@@ -422,6 +422,9 @@ typedef enum _XRefDatabase {
     // Also, if XREF_COUNT > sizeof(uint16), enlarge struct OOVPA.XRefSaveIndex (and Value somehow)
 } XRefDatabase;
 
+// NOTE: XREF_PUBLIC_INDEX is only for library's internal API usage.
+#define XREF_PUBLIC_INDEX(xref_index) xref_index - XREF_KT_COUNT - 1
```
ergo720, line 426: The comment says that this macro should not be used by the client app. However, since the header is in the include directory, it is visible and clients can in fact see and use it. So, I suggest that the macro should be placed in a different header that's only visible internally by the library. If it doesn't exist, then create it.

src/OOVPADatabase/OOVPA.h
```diff
@@ -31,9 +31,44 @@
 // MSVC_EXPAND works around a Visual C++ problem, expanding __VA_ARGS__ incorrectly:
 #define MSVC_EXPAND(x) x
 
+// Wrap C++ style macro function for clang-format to follow our code style.
```
ergo720, line 34:
```suggestion
// Wrap C++ style macro function so that clang-format follows our code style.
```

src/OOVPADatabase/OOVPA.h
```diff
@@ -31,9 +31,44 @@
...
+
+// parentheses
+#define VA_ARGS_EXPAND(...)        __VA_ARGS__
+#define UNPARENTHESES_INVOKE(args) VA_ARGS_EXPAND(args)
+#define UNPARENTHESES(args)        UNPARENTHESES_INVOKE(VA_ARGS_EXPAND args)
+
+// count array type size
+#define COUNT_ARRAYSIZE(type_, array_) (sizeof((type_[])UNPARENTHESES(array_)) / sizeof(type_))
+
+// Increment counter for PARAMS macro to use with.
```
ergo720, line 52:
```suggestion
// Increment counter used by the PARAMS macro.
```

src/OOVPADatabase/OOVPA.h
```diff
@@ -31,9 +31,44 @@
...
 
+// Concatenate two strings into one string for preprocessor to process.
+#define CONCAT(a, b) a##b
+
+// parentheses
+#define VA_ARGS_EXPAND(...)        __VA_ARGS__
+#define UNPARENTHESES_INVOKE(args) VA_ARGS_EXPAND(args)
+#define UNPARENTHESES(args)        UNPARENTHESES_INVOKE(VA_ARGS_EXPAND args)
+
+// count array type size
```
ergo720, line 49:
```suggestion
// get the number of elements of an array
```

src/OOVPADatabase/OOVPA.h
```diff
@@ -200,118 +231,234 @@ typedef enum _eDBScanType {
...
+#define CALL_thi   call_thiscall
+#define CALL_fas   call_fastcall
+#define CALL(Name) CALL_##Name
+
+// For generate symbol's suffix name, mainly for registers, and extend API usage.
+#define PARAM(Param, Name)            Param, Name
+#define PARAM1(Param)                 Param, ""
+#define PARAM_TOKEN_unk(Index, Name)  _unk##Index
+#define PARAM_TOKEN_void(Index, Name) // No argument, do not append to symbol reference.
+#define PARAM_TOKEN_stk2(Index, Name) // Argument is stored in call stack, do not append to symbol reference.
```
ergo720, line 307: Why is there a second stack argument macro if it is identical to the first one below?

RadWolfie: Some parameters like, for example, `LARGE_INTEGER` structure is a size of 8 bytes. So, it pushed double stack, aka 8 bytes (or 64 bits). This is necessary if planning to use header files with `LARGE_INTEGER` structure.

And below should be a dead giveaway of their difference.
https://github.com/Cxbx-Reloaded/XbSymbolDatabase/blob/a167837da6773ca1eeb06b7cef6c88d281d15443/include/libParamTypes.h#L5-L6

If the above doesn't answer to your question, what are you not sure, or thinking, about?

include/libXbSymbolDatabase.h
RadWolfie: This was a typo since Visual Studio's spell checking marked as misspelled.

src/OOVPADatabase/OOVPA.h
ergo720: I understand now, but I believe the name is too vague then. I suggest to rename them to stk32 and stk64 to better clarify their usage/meaning.

include/xref/list_xref.h
RadWolfie: While it's true it is meant for internal library purpose. Although, I had it out here for possible users may want to use it to reduce array size of symbols that are inaccessible via library's APIs. Yet, they have to convert the index back to library's supplied index number.

i.e. unit test tool, but I think it was unnecessary for std map (I think) I used in unit test tool.

src/OOVPADatabase/OOVPA.h
```diff
@@ -200,118 +231,234 @@ typedef enum _eDBScanType {
...
+#define OV_BYTES_8(Offset, Value, ...) OV_BYTE(Offset, Value), MSVC_EXPAND(OV_BYTES_7(Offset + 1, __VA_ARGS__))
+
+#define CALL_unk   call_unknown
+#define CALL_cde   call_cdecl
+#define CALL_std   call_stdcall
+#define CALL_thi   call_thiscall
+#define CALL_fas   call_fastcall
+#define CALL(Name) CALL_##Name
+
+// For generate symbol's suffix name, mainly for registers, and extend API usage.
```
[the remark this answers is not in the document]
RadWolfie: extend is relative to `extend information`, but I'm not sure if it need "ed" at the end for here. 🤷

src/OOVPADatabase/OOVPA.h
RadWolfie: Then that would require replace whole lot of stk and few stk2 inside `PARAM()` argument. 🙁In my opinion, it will also lead some confusion. Like why is there 32 worth of stack pushed onto call stacks when there's only one stack pushed onto call stack? The same for 64 suffix. If we do it this way, then all of the registers will need to follow this way.

Forked ghidra-xbe currently check for stk#, if there is a number at the end. It will be multiply by 4.

include/xref/list_xref.h
ergo720: If it has a valid public usage, then the comment is misleading and should be removed in that case.

src/OOVPADatabase/OOVPA.h
ergo720: What about stk_l and stk_h then, which would stand for low and high respectively (as in low and high 32 bits)? Alternatively, we could also use stk and stk_h to avoid having to replace the many existing stk occurrences in the code.

src/OOVPADatabase/OOVPA.h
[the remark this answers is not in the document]
ergo720: Yes, in that case the extra `ed` is necessary then.

src/OOVPADatabase/OOVPA.h
RadWolfie: What if we found out there are 3, 4, or more pushes onto stack? What would we do about this? For me, it will be stk# to keep things simple and easier to update.

ergo720: As a final suggestion, you could at least update the comments to say the number of the push, as in `stk -> // ... , 1st push` and `stk2 -> // ... , 2nd push`, and so on if any more are encountered in the future.

src/OOVPADatabase/OOVPA.h
```diff
@@ -198,120 +229,236 @@ typedef enum _eDBScanType {
...
+#define CALL_thi   call_thiscall
+#define CALL_fas   call_fastcall
+#define CALL(Name) CALL_##Name
+
+// For generate symbol's suffix name, mainly for registers, and extended API usage.
+#define PARAM(Param, Name)            Param, Name
+#define PARAM1(Param)                 Param, ""
+#define PARAM_TOKEN_unk(Index, Name)  _unk##Index
+#define PARAM_TOKEN_void(Index, Name) // No argument, do not append to symbol reference.
+#define PARAM_TOKEN_stk2(Index, Name) // (Custom) argument is stored in call stack with two stacks, do not append to symbol reference.
```
ergo720, line 307: This comment now makes it sound like there are two stacks, which is not true since every thread has only one stack. It should just say that this is the second push done for an argument passed with the stack.

src/OOVPADatabase/OOVPA.h
RadWolfie: Per discord discussion further, I realized my error of using `stk` terminology when should have used push, or `psh(#)` by team votes. I made a new commit to replace stk(#) with psh(#) since the attempt to fix past commits isn't suitable.

ergo720 (approved):

PR #218, opened 2024-12-09
Partial enhance manual scan system

RadWolfie:
The goal of this pull request is to break away from using groups of OOVPA header files into their own source files. Doing this way will allow LTCG group to have their own respective versioning separated from non-LTCG group.

However, I also notice verification of two symbols are no longer needed as they are integrated into the OOVPA signatures. Which has been removed from this pull request. Plus, a correction of `D3D_g_Stream` reference internally properly reflects to `D3D_g_Stream_i_pVertexBuffer`.


RadWolfie:
Unit test on titles report back with no regressions.

PR #220, opened 2025-07-16
Fix GetTypeInformation_(4|8) to GetTypeInformation(2) wording

### LINKED ISSUE #209: GetTypeInformation_(4|8) wording bug
As I was going through with #208 pull request for XAPI symbol group inspection for parameters... I recently discovered both GetTypeInformation_4 and GetTypeInformation_8 symbols are not standard call functions. Instead they are fastcall functions. So, the parameters are stored in registers instead of onto stack. The pull request doesn't supplied registers onto symbol string for non-LTCG libraries are static generated. A decision will need to be made how to handle this type of situation. Other symbols so far doesn't have this type of conflict at this time.

I tried to look up for the original review remark in Cxbx-Reloaded repository twice but fail to find it again. I'm leaving this open until able to find the original review remark discussion to be link here. Plus to make further decision after re-review the review remark discussion and possible permanent solution for this problem.

RadWolfie:
resolve #209

Since both GetTypeInformation functions do not have arguments stored onto the stack and coexist in at least one XDK build revision(s). Using GetTypeInformation(2) is the best solution to resolve the conflict rather than confusing the GetTypeInformation_(4|8) wording difference.


ergo720:
Could you rename them like this instead -> `GetTypeInformation_4` to `GetTypeInformation_1arg` and `GetTypeInformation_8` to `GetTypeInformation_2arg`? That way, someone would at least have a hint that they are resolving to the 1 and 2 argument variants respectively. Otherwise, with the proposed rename, it's not immediately obvious which is which.

ergo720 (approved):

PR #223, opened 2025-07-27
Minor fixes to unittest and future internal oovpa signatures conflict

RadWolfie:
One commit is forgotten to add to the #221 pull request with the unittest's symbol renamed.

The other commit is mainly to keep OOVPA signatures local only to OOVPA's group files. I am not entirely sure how this was missed duration of untangling from the manual scan process needs to access the OOVPA signatures directly. Relative to #218 pull request.


RadWolfie:
Low risk merge

PR #224, opened 2025-07-29
Fix and update D3DDevice_CopyRects signatures

### LINKED ISSUE #222: BUG: D3D8LTCG's D3DDevice_CopyRects signature(s) isn't accurate
D3DDevice_CopyRects' `1036`, reside in 4627.inl file, signature was detected in:
- Project: Snowblind (ES-025 v3.0, b5849, default.xbe) detected at `0x157c50` but is actually at `0x444920`

Further investigation needed to confirm if `1024` and `1048` signatures may could be wrong as well. However, all of these signatures are in old format and may likely need a complete re-haul.

For possible easier way to find the correct `D3DDevice_CopyRects` function is go into `D3DDevice_PersistDisplay` function then check for call to `D3DDevice_CopyRects`.

RadWolfie:
The original bug was from the Project: Snowblind title received the wrong detected address for D3DDevice_CopyRects. Since the signatures are very old, I went ahead majority replaced them with the new signatures. With these new updated signatures, even more titles are able to detect D3DDevice_CopyRects symbol.

The first commit made these LTCG titles able to detect the symbol:
- NBA 2K2 (Sega Sports) (4039)
- Bruce Lee (4432)
- MURAKUMO (4531)
- Crimson Sea (5120)
- Kung Fu Chaos (5120)
- Midtown Madness 3 (5233)
- Speed Kings (5344)

Then the second commit was brought to my attention relative to the change of order of having allocated stack 2 assembly instructions later, instead of at the beginning. These titles are also LTCG:
- Rally Fusion: Race of Champions (4928)
- Enter The Matrix (5344)
- The Italian Job (5344)
- Star Wars - Battlefront (5849)

Titles with existing symbol detected did not show a change of address or loss of detected symbol.

resolve #222 ticket


PatrickvL (approved):
Only did a visual inspection, but looks good. Just one lint issue to resolve.

RadWolfie:
Lint issue is currently a bug as it should be checking against diffs which isn't working at this time. Please disregard it until macros for wip future support for D3D's member variables are fully integrated. As they are the cause of issue.

src/OOVPADatabase/D3D8/3911.inl
```diff
@@ -5308,17 +5308,22 @@ OOVPA_SIG_MATCH(
...
+    // movzx esi, [ebp + 0x0D]
+    OV_MATCH(0x10, 0x0F, 0xB6, 0x75, 0x0D),
+    // mov bl, [esi + 0x____]
+    OV_MATCH(0x14, 0x8A, 0x9E),
+
+    // Few instructions later will always have
+    // shr e__, 0x3
+    // and e__, 0x7
+    // except above instructions could not be add into signature due
+    // to offset change over time.
```
ergo720, line 5326:
```suggestion
    // except above instructions could not be added into the signature because
    // of the offset changing over time.
```

RadWolfie: done

RadWolfie:
I noticed 1036 (d3d8ltcg) signature was not removed, so I went ahead delete it as part of fixup.

PR #227, opened 2025-08-04
Lower couple of OOVPA signatures

RadWolfie:
MU_Init (5233) signature is detected in "MVP Baseball 2003" title with 5120 build version.

D3DDevice_FlushVertexCache (4134) signature is detected in "Azurik: Rise of Perathia" title with 4039 build version.

Both signatures have been verified.


RadWolfie:
Low risk merge

